PR #7, opened 2021-08-12
ref: Project structure

romtsn:
* Consolidates all apple targets under the `appleMain` sourceset
* Renames the package name (otherwise it clashes with the sentry-java sdk, whose package name is `io.sentry` as well)


build.gradle.kts
```diff
@@ -44,6 +47,9 @@ kotlin {
         }
     }
     ios()
+    watchos()
+    tvos()
+    macosX64()
```
philipphofmann, line 52: Oh yeah 💪

build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
 repositories {
     gradlePluginPortal()
     mavenCentral()
+    google()
```
philipphofmann, line 13: `l`: Just curios. Why do we need the Google repo?

src/jsTest/kotlin/io/sentry/kotlin/multiplatform/SentryTestsJS.kt
```diff
@@ -0,0 +1,5 @@
+package io.sentry.kotlin.multiplatform
+
+class SentryTestsJS {
+
+}
```
philipphofmann, line 5: `l`:
```suggestion

}
```

philipphofmann (approved):
If it works, 🚢

src/jsTest/kotlin/io/sentry/kotlin/multiplatform/SentryTestsJS.kt
romtsn: ktlint needed deadly

build.gradle.kts
romtsn: cause `sentry-android` pulls in some androidx/android deps, and if we don't add it, gradle doesn't know where to pull them from

PR #12, opened 2022-07-13
feat: Add iOS specific extension methods

buenaflor:
This allows the user to capture exceptions and errors directly in swift code and it also serves as a convenience method for the iosMain part of the KMP app. Additionally, the NSException creation has been changed to use the throwable name which is more descriptive.


philipphofmann (approved):
LGTM

PR #13, opened 2022-07-13
feat: Add release process

buenaflor:
Added the necessary config and scripts to run a Sentry release.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -14,4 +14,8 @@ targets:
   - name: github
   - name: registry
     sdks:
-      maven:io.sentry.kotlin.multiplatform:sentry-kotlin-multiplatform:
+      maven:io.sentry:sentry-kotlin-multiplatform:
+      maven:sentry-kotlin-multiplatform-android:
+      maven:sentry-kotlin-multiplatform-iosarm64:
+      maven:sentry-kotlin-multiplatform-iosx64:
+      maven:sentry-kotlin-multiplatform-jvm:
--- a/.craft.yml
+++ b/.craft.yml
@@ -15,7 +15,7 @@ targets:
   - name: registry
     sdks:
       maven:io.sentry:sentry-kotlin-multiplatform:
-      maven:sentry-kotlin-multiplatform-android:
-      maven:sentry-kotlin-multiplatform-iosarm64:
-      maven:sentry-kotlin-multiplatform-iosx64:
-      maven:sentry-kotlin-multiplatform-jvm:
+      maven:io.sentry:sentry-kotlin-multiplatform-android:
+      maven:io.sentry:sentry-kotlin-multiplatform-iosarm64:
+      maven:io.sentry:sentry-kotlin-multiplatform-iosx64:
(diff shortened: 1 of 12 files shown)


.craft.yml
```diff
@@ -0,0 +1,17 @@
...
+    mavenRepoId: ossrh
+    mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
+    android:
+      distDirRegex: /^(sentry-android-|.*-android).*$/
+      fileReplaceeRegex: /\d\.\d\.\d(-\w+(\.\d)?)?(-SNAPSHOT)?/
+      fileReplacerStr: release.aar
+  - name: github
+  - name: registry
+    sdks:
+      maven:io.sentry.kotlin.multiplatform:sentry-kotlin-multiplatform:
```
romtsn, line 17:
```suggestion
      maven:io.sentry:sentry-kotlin-multiplatform:
```

the group should be the same as for the java/android artifacts

romtsn: Also, you would need to create an entry for each target platform, because these will be separate artifacts on maven, like:
```
maven:io.sentry:sentry-kotlin-multiplatform-android:
maven:io.sentry:sentry-kotlin-multiplatform-iosarm32:
maven:io.sentry:sentry-kotlin-multiplatform-iosarm64:
maven:io.sentry:sentry-kotlin-multiplatform-jvm:
// ... and so on
```

romtsn:
this will not work with Craft - currently it will rely on the `distZip` task - you'd need to apply a `Distribution` plugin in the `build.gradle.kts` for that.

Then you'd need to manually wire it up for each platform, take a look here: 
https://github.com/getsentry/sentry-java/blob/main/buildSrc/src/main/java/Publication.kt
and here
https://github.com/getsentry/sentry-java/blob/1c65cc85df87b82f89d9bbab811f444fcfeb0dc5/build.gradle.kts#L101-L147

buenaflor:
@romtsn really appreciate the review! I didn't know if what I did was right at all so it was pretty much a shot in the dark, thanks!

buildSrc/src/main/java/Publication.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+    this.maybeCreate("iosX64").contents {
+        from("build${sep}publications${sep}iosX64")
+        from("build${sep}libs${sep}iosX64")
+        from("build${sep}libs${sep}") {
+            include("sentry-kotlin-multiplatform-iosx64*")
+        }
+    }
+    this.maybeCreate("iosSimulatorArm64").contents {
```
romtsn, line 52: I'm not sure if we need to publish simulator specifically, I guess `iosArm64` would be sufficient already

romtsn (approved):

romtsn:
LGTM, good job 👍 

It's hard for me to check for correctness the artifacts now, but I guess we'll see when we try to publish to maven with Craft and then adjust accordingly

PR #14, opened 2022-07-19
feat: Add tests

buenaflor:
Add e2e test and some simple platform specific tests

I've also implemented a kmp version of SentryId in order to be able to compare against the event id fetched from sentry.io for e2e tests

Apple tests are currently not implemented since there are numerous problems. I will address this in a separate issue


sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryId.kt
```diff
@@ -0,0 +1,35 @@
+package io.sentry.kotlin.multiplatform
+
+class SentryId(private val sentryIdString: String) {
```
marandaneto, line 3: Can't we use the SentryId from the Java SDK here? or because it's java it's not possible?

sentry-kotlin-multiplatform/src/jvmMain/kotlin/io/sentry/kotlin/multiplatform/SentryJvm.kt
```diff
@@ -4,20 +4,22 @@ import io.sentry.Sentry as SentryJvm
...
+        val eventId = SentryJvm.captureException(throwable)
+        return SentryId(eventId.toString())
     }
 
     actual fun close() {
         SentryJvm.close()
     }
 
-}
+}
```
marandaneto, line 25: Always add an empty line at the end of the file, for every file.

PR #20, opened 2022-07-25
feat: Improve nsexceptions transformation

buenaflor:
### Problem

Unhandled exceptions thrown in iOS through Kotlin ends up in a `SIGABRT` exception on sentry.io which doesn't have much useful information as to where the crash happened and how it happened. 

### Solution

Solution: using https://github.com/rickclephas/NSExceptionKt

NSExceptionKt solves this by creating a custom `SentryEvent` and`SentryEnvelope` and storing it as soon as the crash happened. This utilises Kotlin's `setUnhandledExceptionHook`


sentry-kotlin-multiplatform/src/appleTest/kotlin/io/sentry/kotlin/multiplatform/nsexception/CommonAddressesTests.kt
```diff
@@ -0,0 +1,44 @@
...
+        val addresses = listOf<Long>(8, 7, 6, 2, 1, 0)
+        val withoutCommonAddresses = addresses.dropCommonAddresses(commonAddresses)
+        assertEquals(listOf<Long>(8, 7, 6), withoutCommonAddresses)
+    }
+
+    @Test
+    fun testDropCommonEmptyCommon() {
+        val addresses = listOf<Long>(0, 1, 2)
+        val withoutCommonAddresses = addresses.dropCommonAddresses(emptyList())
+        assertSame(addresses, withoutCommonAddresses)
```
marandaneto, line 35: Why not use `assertEquals` here specifically?

sentry-samples/kmp-app/iosApp/iosApp/ContentView.swift
```diff
@@ -4,6 +4,17 @@ import shared
...
 		Text("KPM Sample App " + Platform().platform)
+        VStack() {
+            Button("Capture Message") {
+                SentryKMP().captureMessage(message: "KMP Sample App " + Platform().platform)
+            }
+            Button("Capture Exception") {
+                SharedBusinessLogic().doException()
+            }
+            Button("Hard Crash") {
+                SharedBusinessLogic().hardCrash()
```
marandaneto, line 15: Why not using `SentryKMP`?

sentry-samples/kmp-app/iosApp/iosApp/ContentView.swift
```diff
@@ -4,6 +4,17 @@ import shared
 struct ContentView: View {
 	var body: some View {
 		Text("KPM Sample App " + Platform().platform)
+        VStack() {
+            Button("Capture Message") {
+                SentryKMP().captureMessage(message: "KMP Sample App " + Platform().platform)
```
marandaneto, line 9: `SentryKMP` should be static.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryApple.kt
```diff
@@ -29,6 +34,9 @@ internal actual object SentryBridge {
         val sentryAppleOptions = SentryOptions()
         sentryAppleOptions.dsn = options.dsn
         sentryAppleOptions.attachStacktrace = options.attachStackTrace
+        sentryAppleOptions.beforeSend = { event ->
+            dropKotlinCrashEvent(event as SentryNSExceptionEvent?) as SentryEvent?
```
marandaneto, line 38: I don't follow why we need this: `dropKotlinCrashEvent`

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/SentryUnhandledExceptions.kt
```diff
@@ -0,0 +1,106 @@
...
+ * @see wrapUnhandledExceptionHook
+ */
+fun setSentryUnhandledExceptionHook(): Unit = wrapUnhandledExceptionHook { throwable ->
+    val envelope = throwable.asSentryEnvelope()
+    // The envelope will be persisted, so we can safely terminate afterwards.
+    // https://github.com/getsentry/sentry-cocoa/blob/678172142ac1d10f5ed7978f69d16ab03e801057/Sources/Sentry/SentryClient.m#L409
+    SentrySDK.storeEnvelope(envelope)
+    SentrySDK.configureScope { scope ->
+        scope!!.setTagValue(kotlinCrashedTag, kotlinCrashedTag)
+    }
```
marandaneto, line 42: Would be possible to simply call `SentrySDK.captureEnvelope`?
Or even, instead of wrapping the event in an envelope, rather `SentrySDK.captureEvent`?

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryApple.kt
buenaflor: The SDK automatically captures one crash event on iOS, namely SIGABRT which will be stored as well. We also create a custom event that has the proper stack trace (the primary purpose for this lib). `dropkotlinCrashEvent` is responsible for not sending the [SIGABRT](https://sentry.io/organizations/sentry-sdks/issues/3452869503/?project=5903800&query=is%3Aunresolved) to sentry.io. So without that 2 crash events would be sent if a crash happened, unless it's better that way, what do you think?

sentry-samples/kmp-app/iosApp/iosApp/ContentView.swift
buenaflor: It is static on Kotlin but due to how KMM creates those Obj-C bindings it doesn't seem to be able to create a static callable class in Swift. I'll check if there's a workaround

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/SentryUnhandledExceptions.kt
buenaflor: I tried it and didn't work since this is being called right before the app terminates and `sendEnvelope` being in the background thread: https://github.com/getsentry/sentry-cocoa/blob/c51bd77615c8685569ad3da7bad7a001479ce110/Sources/Sentry/SentryHttpTransport.m#L92

sentry-samples/kmp-app/iosApp/iosApp/ContentView.swift
buenaflor: I wanted the sample to show how exceptions thrown in the shared layer look like (especially the stack trace), but I could also add some more examples using SentryKMP directly

sentry-samples/kmp-app/iosApp/iosApp/ContentView.swift
buenaflor:
```Swift
SentryKMP() === SentryKMP()
```

seems to point to the same instance in Swift, so maybe it's fine to use it like this but document it ?

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/SentryUnhandledExceptions.kt
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Converts `this` [Throwable] to a [SentryEnvelope].
+ */
+private fun Throwable.asSentryEnvelope(): SentryEnvelope {
+    val event = asSentryEvent()
+    val preparedEvent = SentrySDK.currentHub().let { hub ->
+        hub.getClient()?.prepareEvent(event, hub.scope, alwaysAttachStacktrace = false, isCrashEvent = true)
+    } ?: event
+    val item = SentryEnvelopeItem(preparedEvent)
+    val header = SentryEnvelopeHeader(preparedEvent.eventId, SentrySDK.options?.sdkInfo, null)
```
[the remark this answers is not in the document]
buenaflor: Left it as null for now but added a TODO comment for passing traceState when enabling performance monitoring

marandaneto (approved):

PR #22, opened 2022-07-27
feat: add SentryId and use them in capture methods

buenaflor:
It's not possible to use SentryId from either Java or Cocoa SDK since the common (shared) module does not allow platform specific SDKs. 

This SentryId only wraps the UUID as strings


sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt
```diff
@@ -6,10 +6,6 @@ import io.sentry.android.core.SentryAndroid
 import io.sentry.android.core.SentryAndroidOptions
 
 internal actual object SentryBridge {
```
marandaneto, line 8: Since you are modifying the Bridge, the `SentryJvm.kt` and `SentryJs.kt` should also fix at least the method signatures.

sentry-kotlin-multiplatform/src/jsMain/kotlin/io/sentry/kotlin/multiplatform/SentryJs.kt
```diff
@@ -1,15 +1,20 @@
...
     }
 
-    actual fun captureMessage(msg: String) {
+    actual fun captureMessage(message: String): SentryId {
+        return SentryId("")
     }
 
-    actual fun captureException(throwable: Throwable) {
+    actual fun captureException(throwable: Throwable): SentryId {
+        return SentryId("")
```
marandaneto, line 14: In this case, you should add support for `SentryId.EMPTY`, similar to Java.

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryId.kt
```diff
@@ -0,0 +1,44 @@
+package io.sentry.kotlin.multiplatform
+
+class SentryId(private val sentryIdString: String) {
+    var uuidString: String = ""
+        private set
```
marandaneto, line 5: I'd rather prefer to use a `UUID` type so in case `fromStringSentryId` gets an invalid string, that would not work.

buenaflor: UUID is unavailable in this module but I think I can find a workaround

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryId.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    companion object {
+        val EMPTY_ID: SentryId = SentryId("")
```
marandaneto, line 8: That could init directly with `00000000-0000-0000-0000-000000000000`

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryId.kt
marandaneto: Oh I missed that, then its fine, its just a wrapper on top of the other SDKs anyway

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryId.kt
```diff
@@ -0,0 +1,44 @@
...
+    var uuidString: String = ""
+        private set
+
+    companion object {
+        val EMPTY_ID: SentryId = SentryId("00000000-0000-0000-0000-000000000000")
+    }
+
+    init {
+        if (sentryIdString.isEmpty()) {
+            uuidString = "00000000-0000-0000-0000-000000000000"
```
marandaneto, line 13: This could call `EMPTY_ID.toString()` most likely

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryId.kt
```diff
@@ -0,0 +1,44 @@
+package io.sentry.kotlin.multiplatform
+
+class SentryId(private val sentryIdString: String) {
```
marandaneto, line 3: Lets add tests for this class

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryId.kt
marandaneto: Can we generate UUID on Kotlin stdlib? If not, using the `EMPTY_ID` is fine.

buenaflor: I couldn't find a UUID in the stdlib but there's libraries like these ones: https://github.com/benasher44/uuid

buenaflor:
@marandaneto 

I improved the SentryId wrapper by using `expect / actual` (similar to SentryKMP) in order to wrap the specific platform `SentryId` inside the KMP SentryId. It's much nicer this way since we don't have to think about `UUID` anymore and let the platform specific SentryId handle it. And I also added some tests

marandaneto (approved):

PR #23, opened 2022-07-27
ref: Sentry init

buenaflor:
I've played with the inits a bit and found a way to initialize Sentry without having to rely on `Any?` context and renaming `SentryKMP` to `Sentry`

### Result
Android:
```Kotlin
Sentry.init(context) { options ->
  options.dsn = "..."
}
```

iOS:
```Swift
Sentry.start { options in
  options.dsn = "..."
}
```

### Method
Utilizing Kotlin Extensions we can extend the Sentry (KMP) class in the respective platform modules in order to access platform specific initializers

#### Pros:

- No need for `Any?` Context, Android can directly use `Context` as parameter
- No need for overloading
- Naming can be Sentry instead of SentryKMP (no conflict with `io.sentry` package since our package is named `io.sentry.kotlin.multiplatform`
- Android can use `init`, iOS can use `start`

#### Cons:

- init methods are now "scattered" in different Sentry Extension files in the respective platform modules instead of being embedded in SentryKMP and SentryBridge

Is this change worth keeping considering the cons?


sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryExtensions.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Sentry initialization with a context and option configuration handler.
+ *
+ * @param context Application context.
+ * @param configuration Options configuration handler.
+ */
+fun Sentry.init(context: Context, configuration: (SentryKMPOptions) -> Unit) {
+    val options = SentryKMPOptions()
+    configuration.invoke(options)
+    SentryAndroid.init(context, SentryBridge.convertToSentryAndroidOptions(options))
+}
```
marandaneto, line 19: Add new line

sentry-samples/kmp-app/androidApp/src/main/java/sample/kpm_app/android/MainActivity.kt
```diff
@@ -41,7 +42,7 @@ class MainActivity : AppCompatActivity() {
 class SentryApplication : Application() {
     override fun onCreate() {
         super.onCreate()
-        SentryKMP.start(this) {
+        Sentry.init(context = this) {
```
marandaneto, line 45:
```suggestion
        Sentry.init(context) {
```

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryApple.kt
```diff
@@ -6,12 +6,6 @@ import platform.Foundation.NSError
 import platform.Foundation.NSException
 
 internal actual object SentryBridge {
-    actual fun start(context: Any?, configuration: OptionsConfiguration<SentryKMPOptions>) {
```
marandaneto, line 9: Does iOS accept `init` as an extension so it would be called `init` on both SDKs?

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryExtensions.kt
buenaflor: Ahh sorry, I always forget since the ide doesnt do it automatically

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryApple.kt
buenaflor: It's possible to extend it with init but not use it in iOS, it throws a compile error if I try to call Sentry.init because it's a reserved keyword

buenaflor:
@marandaneto 
>`io.sentry.Sentry` on Kotlin most likely. 
> You can import `io.sentry.kotlin.multiplatform.Sentry` on Android too most likely.

It seems that importing `io.sentry.Sentry` is not possible as it gives me an unresolved reference error if I try it

My deps are:
- on KMP SDK: `implementation("io.sentry:sentry-android:6.1.4")`
- the user uses: `api("io.sentry:sentry-kotlin-multiplatform:{version}")`

I can expose `io.sentry` if I use `api` instead of `implementation` for the KMP SDK Android dependency.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryApple.kt
marandaneto: ok gotcha, lets keep `start` then for iOS.

marandaneto (approved):

PR #24, opened 2022-08-01
feat: integrate scope & jvm support

buenaflor:
New Features:

- `configureScope { ... } `
- `captureMessage(message, ScopeCallback)`
- `captureException(throwable, ScopeCallback)`
- JVM support and added Jetpack Compose JVM Desktop to the sample app
- and handful of tests

Scope configuration options: 

- Breadcrumb
- Level
- Context
- Tags
- User
- Extras


sentry-kotlin-multiplatform/build.gradle.kts
```diff
@@ -25,8 +25,7 @@ kotlin {
     android {
         publishAllLibraryVariants()
     }
-
-    jvm()
+    //jvm()
```
marandaneto, line 28: Why is it commented?

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt
```diff
@@ -10,24 +11,31 @@ internal actual object SentryBridge {
         return SentryId(androidSentryId.toString())
     }
 
+    actual fun captureMessage(message: String, scopeCallback: SentryScopeCallback): SentryId {
```
marandaneto, line 14: `scopeCallback` can be an optional parameter with a default value, you dont need 2 overloads.

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt
```diff
@@ -10,24 +11,31 @@ internal actual object SentryBridge {
...
+        val localScope = SentryScope()
+        val scopeConfiguration = localScope.scopeConfiguration(scopeCallback)
```
marandaneto, line 16: I'm not sure this would work, See Java SDK
https://github.com/getsentry/sentry-java/blob/fa3886ffaae668ef2e5d57783df89bfd8bb0c6bb/sentry/src/main/java/io/sentry/Hub.java#L820-L828
First, we have to `clone` the current scope and later apply the callback, is it working this way?

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt
```diff
@@ -1,7 +1,8 @@
 package io.sentry.kotlin.multiplatform
 
 import io.sentry.Sentry
-import io.sentry.android.core.SentryAndroidOptions
+
+private val globalScope = SentryScope()
```
marandaneto, line 5: Why do we need a `globalScope`?

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryBreadcrumb.kt
```diff
@@ -0,0 +1,114 @@
+package io.sentry.kotlin.multiplatform
+
+import io.sentry.Breadcrumb
+import io.sentry.kotlin.multiplatform.extensions.toAndroidSentryLevel
+import io.sentry.kotlin.multiplatform.extensions.toKMPSentryLevel
+
+actual class SentryBreadcrumb {
```
marandaneto, line 7: Is there a reason why we are naming with a `Sentry` prefix?

marandaneto: Ideally, we'd add those data classes into a `protocol` folder.

marandaneto, line 7: Would be possible to be a `data` class? or since its a wrapper that would not make sense?

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/TypeAliases.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import io.sentry.Breadcrumb
+import io.sentry.Scope
+import io.sentry.SentryLevel
+import io.sentry.protocol.User
+
+typealias AndroidSentryLevel = SentryLevel
+typealias AndroidSentryUser = User
+typealias AndroidSentryScope = Scope
+typealias AndroidSentryBreadcrumb = Breadcrumb
```
marandaneto, line 11: Why do we need `typealias`?

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,114 @@
...
+        get() {
+            return scope?.user?.toKMPSentryUser()
+        }
+
+    /**
+     * Initializes this KMP Scope with the Android Scope
+     */
+    internal fun initWithAndroidScope(androidScope: AndroidSentryScope) {
+        this.scope = androidScope
+    }
```
marandaneto, line 34: Should this be a ctor?

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,114 @@
+package io.sentry.kotlin.multiplatform
+
+import io.sentry.ScopeCallback
+import io.sentry.kotlin.multiplatform.extensions.*
+
+actual class SentryScope : ISentryScope {
+
+    // We are directly modifying the Android SDK scope
+    private var scope: AndroidSentryScope? = null
```
marandaneto, line 9: This should not be nullable I guess, See https://github.com/getsentry/sentry-kotlin-multiplatform/pull/24/files#r935254139

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryBreadcrumbExtensions.kt
```diff
@@ -0,0 +1,20 @@
...
+import io.sentry.kotlin.multiplatform.AndroidSentryBreadcrumb
+import io.sentry.kotlin.multiplatform.SentryBreadcrumb
+import io.sentry.util.CollectionUtils
+
+fun SentryBreadcrumb.toAndroidBreadcrumb(): AndroidSentryBreadcrumb {
+    val androidBreadcrumb = AndroidSentryBreadcrumb()
+    androidBreadcrumb.message = this.getMessage()
+    androidBreadcrumb.type = this.getType()
+    androidBreadcrumb.category = this.getCategory()
+    val dataClone = CollectionUtils.newConcurrentHashMap(this.getData() as MutableMap<String?, Any>)
```
marandaneto, line 12: Why do we use `newConcurrentHashMap` if `this.getData()` is never null? and if it is, your cast is going to fail.
The `key` can't be null either.

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryBreadcrumbExtensions.kt
```diff
@@ -0,0 +1,20 @@
...
+    if (dataClone != null) {
+        for ((key, value) in dataClone) {
+            androidBreadcrumb.setData(key!!, value)
```
marandaneto, line 15: `!!` is a bad practice and could lead to crashes.

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryLevelExtensions.kt
```diff
@@ -0,0 +1,26 @@
...
+
+fun SentryLevel.toAndroidSentryLevel(): AndroidSentryLevel {
+    when (this) {
+        SentryLevel.DEBUG -> return AndroidSentryLevel.DEBUG
+        SentryLevel.INFO -> return AndroidSentryLevel.INFO
+        SentryLevel.WARNING -> return AndroidSentryLevel.WARNING
+        SentryLevel.ERROR -> return AndroidSentryLevel.ERROR
+        SentryLevel.FATAL -> return AndroidSentryLevel.FATAL
+    }
+    throw IllegalArgumentException("Sentry Level does not exist")
```
marandaneto, line 14: Never do that, we could cause crashes on people's apps.
We can have sensible defaults or return null.

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryLevelExtensions.kt
```diff
@@ -0,0 +1,26 @@
+package io.sentry.kotlin.multiplatform.extensions
+
+import io.sentry.kotlin.multiplatform.AndroidSentryLevel
+import io.sentry.kotlin.multiplatform.SentryLevel
+
+fun SentryLevel.toAndroidSentryLevel(): AndroidSentryLevel {
```
marandaneto, line 6: I'd rather move those extensions inside of `SentryLevel`

sentry-kotlin-multiplatform/src/androidTest/kotlin/io/sentry/kotlin/multiplatform/SentryBreadcrumbTest.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import io.sentry.kotlin.multiplatform.extensions.toAndroidBreadcrumb
+import io.sentry.kotlin.multiplatform.extensions.toKMPSentryLevel
+import org.junit.Test
+import kotlin.test.assertEquals
+
+class AndroidSentryBreadcrumbTest {
+
+    @Test
+    fun `convert SentryBreadcrumb to AndroidBreadcrumb has proper values`() {
```
marandaneto, line 11: This test can be improved, such as both `level` is null most likely or the default value.
Its important to set values different than the default ones.

sentry-kotlin-multiplatform/src/androidTest/kotlin/io/sentry/kotlin/multiplatform/SentryLevelTest.kt
```diff
@@ -0,0 +1,32 @@
+package io.sentry.kotlin.multiplatform
+
+import org.junit.Test
+import kotlin.test.assertEquals
+import io.sentry.kotlin.multiplatform.extensions.toAndroidSentryLevel
+
+class AndroidSentryLevelTest {
+
+    @Test
+    fun `convert SentryLevel to AndroidSentryLevel has proper value`() {
```
marandaneto, line 10: It's a good practice to have a test for each case instead all of them.
In case one of them fails, you won't know which one failed unless you debug it.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryBreadcrumb.kt
```diff
@@ -0,0 +1,106 @@
+package io.sentry.kotlin.multiplatform
+
+import io.sentry.kotlin.multiplatform.extensions.toCocoaSentryLevel
+import io.sentry.kotlin.multiplatform.extensions.toKMPSentryLevel
+import cocoapods.Sentry.SentryBreadcrumb as CocoaBreadcrumb
+
+actual class SentryBreadcrumb {
```
marandaneto, line 7: That feels like a lot of duplication compared to the other one, the only difference here is that it talks to `CocoaBreadcrumb`.
Maybe we can think a way to generify this.

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,114 @@
...
+        set(value) {
+            scope?.level = value?.toAndroidSentryLevel()
+        }
+        get() {
+            return scope?.level?.toKMPSentryLevel()
+        }
+
+    actual override var user: SentryUser?
+        set(value) {
+            value?.onFieldChanged = {
```
marandaneto, line 21: Why do we need `onFieldChanged`? can we just call it directly?

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,164 @@
...
+        }
+        get() {
+            val userMap = scope?.serialize()?.get("user") as Map<String, String>?
+            if (userMap != null) {
+                val sentryUser = SentryUser()
+                sentryUser.email = userMap["email"] as String
+                sentryUser.username = userMap["username"] as String
+                sentryUser.id = userMap["id"] as String
+                sentryUser.ipAddress = userMap["ip_address"] as String
+                return sentryUser
```
marandaneto, line 42: Those conversions from Map to Object could be extension methods like `SentryUser.fromMap(map): SentryUser`

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,164 @@
...
+    internal fun scopeConfiguration(kmpScopeCallback: SentryScopeCallback): (CocoaScope?) -> Unit {
+        return {
+            initWithCocoaScope(it!!)
+            kmpScopeCallback.run(this)
+        }
+    }
+
+    actual override fun getContexts(): MutableMap<String, Any>? {
+        if (scope?.serialize()?.get("context") != null) {
+            val dict = scope?.serialize()?.get("context") as NSDictionary
```
marandaneto, line 69: Calling a map multiple times has the price of multiple lookups.
Ideally do:
`val context = scope?.serialize()?.get("context")`
And reuse the context var only.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,164 @@
...
+    actual override fun addBreadcrumb(breadcrumb: SentryBreadcrumb) {
+        scope?.addBreadcrumb(breadcrumb.toCocoaBreadcrumb())
+    }
+
+    actual override fun clearBreadcrumbs() {
+        scope?.clearBreadcrumbs()
+    }
+
+    private fun setContextForAnyValue(key: String, value: Any) {
+        val map = HashMap<Any?, Any>()
```
marandaneto, line 102: You can use the Kotlin built-in functions for defining a map.
https://www.geeksforgeeks.org/kotlin-mutablemapof/
The keys are always `String` iirc.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,164 @@
...
+        map.put("value", value)
```
marandaneto, line 103: Only JSON primitive types should be wrapped in the `value` object.
eg.
`{ "key": "myValue" }` would not work because the `value` should be an object, in this case we wrap it.
`{ "key": { "value": "myValue" } }`

marandaneto: We just need to be sure that not every type here is wrapped into the `value` object.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
```diff
@@ -0,0 +1,164 @@
...
+        scope?.setExtraValue(value, key)
+    }
+
+    actual override fun removeExtra(key: String) {
+        scope?.removeExtraForKey(key)
+    }
+
+    actual override fun clear() {
+        user = null
+        level = null
```
marandaneto, line 161: Why do we need that, isn't the state on Native only?

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryBreadcrumbExtensions.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.sentry.kotlin.multiplatform.CocoaSentryBreadcrumb
+import io.sentry.kotlin.multiplatform.SentryBreadcrumb
+
+fun SentryBreadcrumb.toCocoaBreadcrumb(): CocoaSentryBreadcrumb {
+    val cocoaBreadcrumb = CocoaSentryBreadcrumb()
+    this.getMessage().let { cocoaBreadcrumb.setMessage(it) }
+    this.getCategory().toString().let { cocoaBreadcrumb.setCategory(it) }
+
+    this.getData().let {
+        if (it != null) {
```
marandaneto, line 12: If you use `?.` you don't need to `it != null`

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryLevelExtensions.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlinx.cinterop.convert
+import platform.darwin.NSUInteger
+import cocoapods.Sentry.SentryLevel as CocoaSentryLevel
+
+//    kSentryLevelNone = 0,
+//    kSentryLevelDebug = 1,
+//    kSentryLevelInfo = 2,
+//    kSentryLevelWarning = 3,
+//    kSentryLevelError = 4,
+//    kSentryLevelFatal = 5
```
marandaneto, line 13: Maybe we use constants here, so we dont hard code the numbers.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryLevelExtensions.kt
```diff
@@ -0,0 +1,66 @@
...
+        else -> {
+            level = 0
+            return level.convert()
+        }
+    }
+}
+
+fun CocoaSentryLevel.toKMPSentryLevel(): SentryLevel? {
+    val level = this
+    if (level.convert<Int>().compareTo(1) == 0) {
```
marandaneto, line 47: I guess we can simplify this, it seems a bit complicated to a simple int comparison.

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtension.kt
```diff
@@ -0,0 +1,16 @@
+package io.sentry.kotlin.multiplatform.extensions
+
+import cocoapods.Sentry.SentryEvent
+import io.sentry.kotlin.multiplatform.CocoaSentryOptions
+import io.sentry.kotlin.multiplatform.nsexception.dropKotlinCrashEvent
+import NSException.Sentry.SentryEvent as NSExceptionSentryEvent
+
+fun io.sentry.kotlin.multiplatform.SentryOptions.toCocoaSentryOptions(): CocoaSentryOptions {
```
marandaneto, line 8:
```suggestion
fun SentryOptions.toCocoaSentryOptions(): CocoaSentryOptions {
```

We can import the package and use the class name directly

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryUser.kt
```diff
@@ -0,0 +1,64 @@
...
+    var ipAddress: String?
+
+    /**
+     * Additional arbitrary fields, as stored in the database (and sometimes as sent by clients). All
+     * data from `self.other` should end up here after store normalization.
+     */
+    var other: Map<String, String>
+
+    /** Unknown fields, only internal usage. */
+    var unknown: Map<String, Any>
```
marandaneto, line 63: Maybe they should be all mutable?

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryBreadcrumbFactory.kt
```diff
@@ -0,0 +1,111 @@
+package io.sentry.kotlin.multiplatform
+
+/**
+ * The factory is responsible for creating Breadcrumb objects in SentryBreadcrumb
+ * in order to avoid code duplication.
+ *
+ * This is only for internal usage and is not exposed to the user.
+ */
+internal object SentryBreadcrumbFactory {
```
marandaneto, line 9: Wouldn't be possible to just call the Breadcrumb class from Native and convert it back to the KMP type?

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryScopeCallback.kt
```diff
@@ -0,0 +1,5 @@
+package io.sentry.kotlin.multiplatform
+
+fun interface SentryScopeCallback {
+    fun run(scope: SentryScope)
+}
```
marandaneto, line 5: Missing new line

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryScopeCallback.kt
```diff
@@ -0,0 +1,5 @@
+package io.sentry.kotlin.multiplatform
+
+fun interface SentryScopeCallback {
```
marandaneto, line 3: Do we need to declare an interface?
Maybe we could use higher-order functions/callbacks.

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt
buenaflor: This was my first iteration, now that we are directly modifying the scope we don't need it anymore

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/TypeAliases.kt
buenaflor: This makes it easier to work in classes with different types from KMP and Android for example so you immediately know which type you are dealing with. I have been importing those with `as [name]` alot and wanted some consistency. from `SentryLevel and SentryLevel` to `SentryLevel and AndroidSentryLevel`. But it may be good to omit the `Sentry` in the name though

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
buenaflor: It can be, but the KMP Scope creation has to be put inside the `configureScope` callback so we can access the Platform Scope. Shouldn't be a problem though I think

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt
buenaflor: if for example the user does something like this:

```Kotlin
configureScope { scope ->
  scope.user = User()
  scope.user.username = "username"
}

Then the set(value) is not triggered during 'user.username = ' so I put in a callback to `SentryUser` fields so we know when to update the scope's user. Maybe there's a much better way of doing this? maybe pass the scope to User and modify it directly there?

File: sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtension.kt (hunk 25 above)
Comment (buenaflor) [reply sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtension.kt:8]: Oh thanks I didn't even notice that

File: sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryScopeCallback.kt (hunk 29 above)
Comment (buenaflor) [reply sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryScopeCallback.kt:3]: I'll replace it, it's actually better if we use higher-order functions since Kotlin - Obj-C conversion doesn't recognize these interfaces as callbacks

File: sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt (hunk 3 above)
Comment (buenaflor) [reply sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt:16]: Isn't this being done internally since we modify the scope directly? When I tested it the scope was independently isolated from others only for that event but I'll have to check again

File: sentry-kotlin-multiplatform/build.gradle.kts (hunk 1 above)
Comment (buenaflor) [reply sentry-kotlin-multiplatform/build.gradle.kts:28]: the expect/actual keywords force me to write implementations for every specified target and it's much easier to focus on iOS and Android if the compiler doesn't send out an error all the time for missing implementations.

Also since there is a lot of overlap between JVM and Android sdks I want to explore a way to reduce code duplication in another issue if there's time left

File: sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt (hunk 2 above)
Comment (buenaflor) [reply sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryAndroid.kt:14]: Kotlin -> Obj-C is sometimes ugly so default parameters are pretty much nonexistent

https://github.com/getsentry/sentry-kotlin-multiplatform/blob/main/sentry-samples/kmp-app/iosApp/iosApp/ContentView.swift#L15

iOS users would need to write it like this: `captureMessage("message: "...", scopeCallback = nil)

File: sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryLevelExtensions.kt (hunk 12 above)
Comment (buenaflor) [reply sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryLevelExtensions.kt:6]: There is no SentryLevel on Android or iOS modules (it's only an enum) so I can't move them elsewhere

File: sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt (hunk 16 above)
Comment (marandaneto) [reply sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt:23]: the goal is to do:
`val myUser = scope.user`
`scope.user = myUser`

File: sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryScope.kt [hunk 30]
```diff
@@ -0,0 +1,116 @@
+package io.sentry.kotlin.multiplatform
+
+import io.sentry.ScopeCallback
+import io.sentry.kotlin.multiplatform.extensions.*
+import io.sentry.kotlin.multiplatform.protocol.SentryBreadcrumb
+import io.sentry.kotlin.multiplatform.protocol.SentryUser
+
+actual class SentryScope : ISentryScope {
```
marandaneto, line 8:
```suggestion
actual class SentryScope(scope: ISentryScope) : ISentryScope {
```

buenaflor:
@marandaneto I've refactored a lot of things, hopefully it looks a bit better now. I've also tested the functionality manually and it works so far on Android and iOS (I also added more tests)

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/ScopeCocoaImpl.kt
```diff
@@ -0,0 +1,126 @@
...
+            return level?.let { SentryLevel.valueOf((it as String).uppercase()) }
+        }
+
+    override var user: User?
+        set(value) {
+            scope.setUser(value?.toCocoaUser())
+        }
+        get() {
+            val map = scope.serialize()["user"] as? Map<String, String>?
+            return map?.let { User.fromMap(map) }
```
brustolin, line 26: m: This is probably so slow.
Can't we get the user property directly?

buenaflor: Cocoa sdk doesn't have public getters for the properties in scope, only `setUser()`. I'll open a feature request for that soon

sentry-samples/kmp-app/desktopApp/src/jvmMain/kotlin/Main.kt
```diff
@@ -0,0 +1,66 @@
...
+    Window(onCloseRequest = ::exitApplication) {
+        Sentry.init(optionsConfiguration())
+
+        // Shared scope across all platforms
+        configureSharedScope()
+
+        // Add platform specific scope in addition to the shared scope
+        Sentry.configureScope {
+            it.setContext("Android Context", mapOf("context1" to 12, "context2" to false))
+            it.addBreadcrumb(Breadcrumb.debug("initialized Sentry on Android"))
```
brustolin, line 61: Is this an Android app?

buenaflor: Seems like I forgot to change it after refactoring, it should be JVM not Android thx

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/Scope.kt
```diff
@@ -0,0 +1,216 @@
+package io.sentry.kotlin.multiplatform
+
+import io.sentry.kotlin.multiplatform.protocol.Breadcrumb
+import io.sentry.kotlin.multiplatform.protocol.User
+
+class Scope constructor(private val scope: ISentryScope) : ISentryScope {
```
brustolin, line 6: h: No attachments for KMP scope?

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/ScopeCocoaImpl.kt
brustolin: I dont think we want to add this to cocoa SDK. But, we do have a private header (is like an internal thing) with this and other properties. 
Can KVM add this private headers in the cocoa bridge?

brustolin (approved):
Codewise it looks good. 
Im not familiar with KMP, so I can't say whether all configurations looks good.

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/Scope.kt
buenaflor: I will add more features in the next couple days/weeks, the plan was to introduce some basic Scope options first

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/ScopeCocoaImpl.kt
buenaflor: hm good point, yes I think there's a possibility to create a bridge to get those, I'll investigate that

buenaflor: @brustolin I added custom headers to access the properties, no need for serialize now

PR #29, opened 2022-08-09
feat: js support

buenaflor:
This is just a draft so some things are still wonky.

The current goal is to support Kotlin code usage in Javascript. I've attached a simple vanilla JS Browser App that calls Kotlin code (`Sentry.init` and so on).

A lot of stuff works: capturing messages, capturing exceptions, setting scopes, context, tags, etc but the Stacktrace references addresses located in the webpack internals. The question is how to represent the stacktrace since Kotlin is compiled to Javascript. At least it shows where the function has been called in the HTML and which functions have caused the exception. Also there are errors regarding missing source code. 

Screenshot 2022-08-09 at 07 18 06


AbhiPrasad:
> The question is how to represent the stacktrace since Kotlin is compiled to Javascript. At least it shows where the function has been called in the HTML and which functions have caused the exception. Also there are errors regarding missing source code.

We'll need to upload source maps to make this happen. Seems like [we can generate them using a gradle plugin](https://kotlinlang.org/docs/js-debugging.html)? We can then upload them to the server using `sentry-cli`. See: https://docs.sentry.io/product/cli/releases/#managing-release-artifacts

PR #31, opened 2022-08-10
ref: improve Sentry Options

buenaflor:
This introduces some additions to the options and refactoring. Refactoring the sourceSets is needed due to the different initialization options. For example: Screenshots are only available on Android and iOS but not on JVM, watchOS, macOS, tvOS. So we need to group those together.

In total there will be 4 different initializers internally: JVM, Android, iOS and the rest of Cocoa targets (bundled into one common sourceSet)

With this PR it's also possible to have a shared initializer
Untitled drawio (1)


sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryInit.kt
```diff
@@ -0,0 +1,16 @@
+package io.sentry.kotlin.multiplatform
+
+import android.content.Context
+import io.sentry.android.core.SentryAndroid
+import io.sentry.kotlin.multiplatform.extensions.toAndroidSentryOptionsCallback
+
+actual fun initJvmTarget(context: Context?, configuration: OptionsConfiguration) {
```
marandaneto, line 7: Should this method be renamed to initAndroidTarget?

marandaneto: `Context` cannot be nullable.

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryKMP.kt
```diff
@@ -4,10 +4,33 @@ import io.sentry.kotlin.multiplatform.protocol.SentryId
 import io.sentry.kotlin.multiplatform.protocol.UserFeedback
 
 typealias ScopeCallback = (Scope) -> Unit
+typealias OptionsConfiguration = (SentryOptions) -> Unit
+
+expect abstract class Context
```
marandaneto, line 9: Why do we need that?

sentry-samples/kmp-app/androidApp/src/main/AndroidManifest.xml
```diff
@@ -8,6 +8,7 @@
         android:theme="@style/AppTheme"
         android:name=".SentryApplication">
         <meta-data android:name="io.sentry.auto-init" android:value="false" />
+        <meta-data android:name="io.sentry.attach-screenshot" android:value="true" />
```
marandaneto, line 11: I'd rather set that on the options directly instead of the Manifest, just for the sake of simplicity for now.

sentry-samples/kmp-app/desktopApp/src/jvmMain/kotlin/Main.kt
```diff
@@ -13,13 +13,8 @@ import androidx.compose.ui.graphics.Color
...
 import androidx.compose.ui.window.application
 import io.sentry.kotlin.multiplatform.Sentry
-import io.sentry.kotlin.multiplatform.SentryOptions
 import io.sentry.kotlin.multiplatform.protocol.Breadcrumb
-import io.sentry.kotlin.multiplatform.protocol.init
-import sample.kpm_app.LoginImpl
-import sample.kpm_app.Platform
-import sample.kpm_app.configureSharedScope
-import sample.kpm_app.optionsConfiguration
+import sample.kpm_app.*
```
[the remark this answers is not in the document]
buenaflor: yes, I'll merge the ktlint pr and format it

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryInit.kt
buenaflor: I don't think I can change a function name with expect/actual, expect defines the function that the submodules have to implement with actual and the names and signatures have to match.

both jvmMain and androidMain implement actual fun initJvmTarget, maybe I can rename it to something more generic?

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryKMP.kt
buenaflor: this is a rather 'elegant' I'd say workaround to avoid `Any` context.

`expect abstract class` expects us to implement this in every submodule.

so on androidMain we say `actual typealias Context = Context (android context)`

on Cocoa it's just an empty class: `actual abstract class Context`

This works because the expected class Context doesn't have a body.

Now if you try to initialize the Android SDK, you can only pass in the Android Context and nothing else which is much better than using `Any`

sentry-samples/kmp-app/shared/src/commonMain/kotlin/sample/kpm_app/AppSetup.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+/**
+ * Initializes Sentry with given options.
+ * Make sure to hook this into your native platforms as early as possible
+ */
+fun initializeSentry(context: Context) {
+    Sentry.init(context, optionsConfiguration())
+}
```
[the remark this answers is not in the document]
buenaflor: This is the shared initializer which is always `Sentry`

Depending on where you call it, SDK will route to the correct implementation, so calling `initializeSentry` on Android will eventually end up calling `SentryAndroid.init(...)` in the SDK. 

Like in the graph, KMP chooses the path to a correct leaf node

README.md
```diff
@@ -75,15 +84,60 @@ class YourApplication : Application() {
...
+    ) -> Bool {
+        Sentry.shared.doInit() { options in
+          options.dsn = "__DSN__"
+        }
+        return true        
+    }
+}
+```
+
+## Shared Initializer
```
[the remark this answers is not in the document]
buenaflor: Yes this is something the user needs to do if they want to use a shared initializer, it's an extra step but if they don't want to do that they can also just initialize directly without doing anything extra. But I think it's also a good thing to document how to do it since code sharing is the main concept in KMP.

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryInit.kt
buenaflor: A non nullable Context is hard to implement since multiple platforms share initializer signatures and it's hard to force users only to use one specific initializer. For example it's possible on Android to use  `Sentry.init { ... }` which is incorrect

README.md
```diff
@@ -60,30 +60,86 @@ cocoapods {
...
```
marandaneto, line 105: Lets move this up since this is the preferred option.

marandaneto (approved):

PR #33, opened 2022-08-13
build: add formatting with ktlint

buenaflor:
This uses the spotless plugin with ktlint to check and apply formats.

Added `format.yml` to trigger on pull requests similar to: https://github.com/getsentry/sentry-java/blob/main/.github/workflows/format-code.yml

Changes in the codebase have been made by the formatter


marandaneto:
@buenaflor consider adding [swift lint](https://github.com/getsentry/sentry-dart/blob/1998016942baf8573e813631288f7e6534394895/.github/workflows/flutter.yml#L182-L189) for the iOS bits.
you can create a new issue for that instead of addressing it right now.

marandaneto (approved):

PR #37, opened 2022-08-17
fix craft.yml and add CHANGELOG

buenaflor:
Add missing changelog and fix craft yml artifacts setting


buenaflor:
@marandaneto I'm not sure if the Android config is needed in the craft yml since the distribution of the sentry-kotlin-multiplatform module only contains jars:
Screenshot 2022-08-17 at 08 29 05

marandaneto:
> @marandaneto I'm not sure if the Android config is needed in the craft yml since the distribution of the sentry-kotlin-multiplatform module only contains jars: Screenshot 2022-08-17 at 08 29 05

Likely not, all good then, let's remove it.
I hope users don't need to add all those dependencies then.
Strange is that there's only a single `pom` file. what's the content of it? usually each jar should have its own `pom` to be published, wondering if this is going to work.

buenaflor:
> > @marandaneto I'm not sure if the Android config is needed in the craft yml since the distribution of the sentry-kotlin-multiplatform module only contains jars: Screenshot 2022-08-17 at 08 29 05
> 
> Likely not, all good then, let's remove it. I hope users don't need to add all those dependencies then. Strange is that there's only a single `pom` file. what's the content of it? usually each jar should have its own `pom` to be published, wondering if this is going to work.

I think there are too many unused dependencies in there. For example this is what `./gradlew publishToMavenLocal` produces which works locally with `api(io.sentry:sentry-kotlin-multiplatform:0.0.1`

Screenshot 2022-08-17 at 08 52 52

marandaneto:
@buenaflor can you place the content of the `pom` file here? I wanna check it out.

romtsn:
We still have to publish all of the artifacts for all platforms. The `sentry-kotlin-multiplatform-0.0.1.jar` is just a metadata jar. For example, check how sqldelight is published (e.g. the `runtime` artifacts) - https://repo.maven.apache.org/maven2/app/cash/sqldelight/. There's a full publication for each platform (jar, sources, javadoc and pom).

For android distribution it's still gonna be an `aar`, it might be just in a different folder when you run `publishToMavenLocal`, so the android config is still necessary. (because we have AndroidManifest)

romtsn:
Also, I noticed that the source dirs should probably fixed - e.g. look at `commonJvmMain` - there's a `io.sentry.kotlin.multiplatform` folder with dots instead of slashes, so I'm not sure if it's recognized as a source folder?

image

buenaflor:
@marandaneto 

```
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <groupId>io.sentry</groupId>
  <artifactId>sentry-kotlin-multiplatform</artifactId>
  <version>0.0.1</version>
  <name>Sentry Kotlin Multiplatform SDK</name>
  <description>SDK for sentry.io</description>
  <url>https://github.com/getsentry/sentry-kotlin-multiplatform</url>
  <licenses>
    <license>
      <name>MIT</name>
      <url>http://www.opensource.org/licenses/mit-license.php</url>
    </license>
  </licenses>
  <developers>
    <developer>
      <id>getsentry</id>
      <name>Sentry Team and Contributors</name>
      <url>https://github.com/getsentry/</url>
    </developer>
  </developers>
  <scm>
    <connection>scm:git:git://github.com/getsentry/sentry-kotlin-multiplatform.git</connection>
    <developerConnection>scm:git:ssh://git@github.com/getsentry/sentry-kotlin-multiplatform.git</developerConnection>
    <url>https://github.com/getsentry/sentry-kotlin-multiplatform</url>
  </scm>
  <dependencies>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib-jdk8</artifactId>
      <version>1.7.0</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib</artifactId>
      <version>1.7.0</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>
</project>
```

buenaflor:
@romtsn yeah, I just tried running it on mavenLocal without the other platforms and it didn't work, but I still have too many unnecessary jar files in the sentry-kotlin-multiplatform artifact right? do I need to match the distribution with what mavenLocal publishes?

marandaneto:
> @romtsn yeah, I just tried running it on mavenLocal without the other platforms and it didn't work, but I still have too many unnecessary jar files in the sentry-kotlin-multiplatform artifact right? do I need to match the distribution with what mavenLocal publishes?

I'd say yes most likely, all the files that end up in the `mavenLocal` folder, are the ones that need to be published.

marandaneto:
Marking as a draft, let me know once you figure this out for review.

buenaflor:
@marandaneto should be ready now for review

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+# Changelog
+
+## Unreleased
+
+### Features
+ - JVM, Android, iOS, macOS, watchOS, tvOS integration
```
marandaneto, line 6:
```suggestion
### Features

 - JVM, Android, iOS, macOS, watchOS, tvOS integration
```

marandaneto (approved):

Issue #44, opened 2023-01-01
commonizeCInterop fails with `Unresolved classifier: platform/darwin/xpc_rich_error_t` on XCode 14.2


erksch:
### Environment

Apple M1 Pro
XCode 14.2
macOS Ventura 13.0.1

### Steps to Reproduce

```
./gradlew build
```

### Expected Result

Builds successfully.

### Actual Result

```
> Task :sentry-kotlin-multiplatform:commonizeCInterop FAILED
Exception in thread "main" java.lang.IllegalStateException: Unresolved classifier: platform/darwin/xpc_rich_error_t
        at org.jetbrains.kotlin.commonizer.metadata.CirTypeAliasExpansion$Companion.create(CirTypeAliasExpander.kt:183)
        at org.jetbrains.kotlin.commonizer.metadata.CirDeserializers.type(CirDeserializers.kt:367)
        at org.jetbrains.kotlin.commonizer.metadata.CirDeserializers.valueParameter(CirDeserializers.kt:199)
        at org.jetbrains.kotlin.commonizer.metadata.CirDeserializers.function(CirDeserializers.kt:178)
        at org.jetbrains.kotlin.commonizer.tree.deserializer.CirTreeFunctionDeserializer.invoke(CirTreeFunctionDeserializer.kt:29)
        at org.jetbrains.kotlin.commonizer.tree.deserializer.CirTreePackageDeserializer.invoke(CirTreePackageDeserializer.kt:33)
        at org.jetbrains.kotlin.commonizer.tree.deserializer.CirTreeModuleDeserializer.invoke(CirTreeModuleDeserializer.kt:32)
        at org.jetbrains.kotlin.commonizer.tree.deserializer.RootCirTreeDeserializer.invoke(RootCirTreeDeserializer.kt:37)
 ...
```

Seems to be related to this issue: https://youtrack.jetbrains.com/issue/KT-54225/Native-update-to-Xcode-14.1

Updating Kotlin to `1.8.0` (Kotlin and Kotlin Gradle Plugin) solves this issue but creates a new one.

```
> Task :sentry-samples:kmp-app:shared:commonizeCInterop FAILED
Exception in thread "main" java.lang.IllegalStateException: Unresolved classifier: cocoapods/Sentry/SentryBreadcrumb
	at org.jetbrains.kotlin.commonizer.metadata.CirTypeAliasExpander.expandClassType(CirTypeAliasExpander.kt:216)
[...]

PR #48, opened 2023-02-28
fix: build and bump versions

buenaflor:
this PR should fix most issues regarding the build and bump versions of Kotlin and Multiplatform especially #44

additionally, I confirmed the correctness of the artifacts by publishing them on [mymavenrepo](https://mymavenrepo.com/) and testing the SDK in sample apps. Here is the published repo: https://mymavenrepo.com/repo/IxbzFmjbPjBH4llKSvgl/

testing can be done through 
 - adding the repository: `maven {
            url = uri("https://mymavenrepo.com/repo/IxbzFmjbPjBH4llKSvgl/")
        }`
 - importing the dependency in gradle for example: `api("io.sentry:sentry-kotlin-multiplatform:0.0.1")`

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -9,5 +9,7 @@ repositories {
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
 }
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -10,6 +10,6 @@ repositories {
 
 tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions {
-        jvmTarget = "1.8"
+        jvmTarget = properties["jvm.version"].toString()
     }
 }
--- a/sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/Attachment.kt
+++ b/sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/Attachment.kt
@@ -17,7 +17,7 @@ actual class Attachment : IAttachment {
         get() = cocoaAttachment.data?.toByteArray()
 
     actual override val contentType: String?
-        get() = cocoaAttachment.contentType
+        get() = cocoaAttachment.contentType ?: "application/octet-stream"
 
     actual companion object {
         actual fun fromScreenshot(screenshotBytes: ByteArray): Attachment {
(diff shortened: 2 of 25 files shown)


buildSrc/build.gradle.kts
```diff
@@ -9,5 +9,7 @@ repositories {
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
+    kotlinOptions {
+        jvmTarget = "1.8"
```
marandaneto, line 13: We can add to the Config file as well

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/Attachment.kt
```diff
@@ -17,7 +17,7 @@ actual class Attachment : IAttachment {
         get() = cocoaAttachment.data?.toByteArray()
 
     actual override val contentType: String?
-        get() = cocoaAttachment.contentType
+        get() = cocoaAttachment.contentType ?: "application/octet-stream"
```
marandaneto, line 20: Is there a reason to fallback? if the field is nullable and the server has its own default

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtensions.kt
```diff
@@ -19,7 +19,9 @@ internal fun CocoaSentryOptions.applyCocoaBaseOptions(options: SentryOptions) {
     this.dsn = options.dsn
     this.attachStacktrace = options.attachStackTrace
     this.dist = options.dist
-    this.environment = options.environment
+    options.environment?.let {
+        this.environment = it
+    }
```
marandaneto, line 24: Seems like the old way achieves the same thing.

marandaneto (approved):
Left a few comments but overall LGTM

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtensions.kt
buenaflor: this.environment changed from nullable to requiring a non-null value

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/Attachment.kt
buenaflor: If it's enforced in the server then it's not needed. I changed it because my attachment tests were failing but I'll adjust those as well

marandaneto: I dont think that would be the problem, but double-check if its still failing.

sentry-kotlin-multiplatform/src/commonTest/kotlin/io/sentry/kotlin/multiplatform/AttachmentTest.kt
```diff
@@ -6,7 +6,7 @@ import kotlin.test.assertEquals
 
 class AttachmentTest {
 
-    private val defaultContentType = "application/octet-stream"
+    private val defaultContentType = null
```
marandaneto, line 9: Maybe just delete `defaultContentType` altogether? since its not needed

PR #50, opened 2023-03-03
feat: override sdk name and version

buenaflor:
**Overrides the sdk name and version in the specific SDKs**

example: 
 sdk name: `sentry.kotlin.kmp`
 sdk version: `0.0.1` (currently)

packages depending on which platform the user uses:
jvm: 
 - `name: "maven:io.sentry:sentry", version: "6.14.0"
 
android:
 - `name: "maven:io.sentry:sentry-android", version: "6.14.0"

apple:
 - `name: "cocoapods:sentry-cocoa", version: "8.2.0"

LMK if some naming is incorrect 

**Testing:**
The [sentry events](https://sentry-sdks.sentry.io/issues/3961805714/?project=5903800&referrer=project-issue-stream) now show the specified sdk name and version at the bottom. Also tested it by temporarily modifying a Look on Looker: 

Screenshot 2023-03-06 at 09 43 58


gradle.properties
```diff
@@ -17,13 +17,8 @@ versionName=0.0.1
 # Increase memory for in-process compiler execution.
 org.gradle.jvmargs=-Xmx3g
 
-# Because of: Please try to disable compiler caches and rerun the build. To disable compiler caches, add the following line to the gradle.properties file in the project's root directory
-kotlin.native.cacheKind=none
-
```
buenaflor, line 22: Disabling cache for kotlin/native was used before to fix build issues but it can be safely removed now

gradle.properties
```diff
@@ -17,13 +17,8 @@ versionName=0.0.1
...
 # https://kotlinlang.org/docs/migrating-multiplatform-project-to-14.html#migrate-to-the-hierarchical-project-structure
-kotlin.mpp.enableGranularSourceSetsMetadata=true
 kotlin.mpp.enableCompatibilityMetadataVariant=true
-kotlin.native.enableDependencyPropagation=false
```
buenaflor, line 26: removing these two options is safe since kmp does them by default

build.gradle.kts
```diff
@@ -9,10 +9,11 @@ plugins {
     kotlin(Config.cocoapods).version(Config.kotlinVersion).apply(false)
     id(Config.jetpackCompose).version(Config.composeVersion).apply(false)
     id(Config.androidGradle).version(Config.agpVersion).apply(false)
+    id(Config.BuildPlugins.buildConfig).version(Config.BuildPlugins.buildConfigVersion).apply(false)
```
marandaneto, line 12: Indentation here is off

buildSrc/src/main/java/Config.kt
```diff
@@ -46,4 +51,14 @@ object Config {
         val tvosDeploymentTarget = "11.0"
         val watchosDeploymentTarget = "4.0"
     }
+
+    object Sentry {
+        val kotlinMultiplatformSdkName = "sentry.kotlin.kotlin-multiplatform"
```
marandaneto, line 56: Not sure if this is correct.
I'd double check our docs about the SDK naming

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryAndroidOptionsExtensions.kt
```diff
@@ -9,4 +10,9 @@ internal fun SentryOptions.toAndroidSentryOptionsCallback(): (SentryAndroidOptio
 
     // Apply Android specific options
     it.isAttachScreenshot = this.attachScreenshot
+    it.sdkVersion = sdk.toJvmSdkVersion()
+    it.sdkVersion?.addPackage(
+        BuildKonfig.SENTRY_ANDROID_PACKAGE_NAME,
+        BuildKonfig.SENTRY_ANDROID_VERSION
+    )
```
marandaneto, line 17: Can `it.sdkVersion` be null here? or can `sdk.toJvmSdkVersion()` return null?

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtensions.kt
```diff
@@ -28,11 +30,20 @@ internal fun CocoaSentryOptions.applyCocoaBaseOptions(options: SentryOptions) {
     this.enableAutoSessionTracking = options.enableAutoSessionTracking
     this.beforeSend = { event ->
         dropKotlinCrashEvent(event as NSExceptionSentryEvent?) as SentryEvent?
+
+        val cocoaName = BuildKonfig.SENTRY_COCOA_PACKAGE_NAME
+        val cocoaVersion = BuildKonfig.SENTRY_COCOA_VERSION
+        options.sdk.apply {
+            val pkg = Package(cocoaName, cocoaVersion)
+            if (!this.packages.contains(pkg)) {
```
marandaneto, line 38: I don't think this would work because `contains` is checking a reference of `pkg` here.
You'd need to match the `cocoaName` most likely.

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtensions.kt
```diff
@@ -28,11 +30,20 @@ internal fun CocoaSentryOptions.applyCocoaBaseOptions(options: SentryOptions) {
...
+                this.packages.add(Package(cocoaName, cocoaVersion))
```
marandaneto, line 39:
```suggestion
                this.packages.add(pkg)
```

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/protocol/SdkVersion.kt
```diff
@@ -0,0 +1,20 @@
+package io.sentry.kotlin.multiplatform.protocol
+
+actual data class SdkVersion actual constructor(
+    actual val name: String,
+    actual val version: String,
+    actual val packages: MutableList<Package>
+) {
+    fun toCocoaSdkVersion(): Map<Any?, *> {
```
marandaneto, line 8:
```suggestion
    fun toCocoaSdkVersion(): Map<String, Any> {
```

We can be more specific here

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/protocol/SdkVersion.kt
```diff
@@ -0,0 +1,20 @@
+package io.sentry.kotlin.multiplatform.protocol
+
+actual data class SdkVersion actual constructor(
+    actual val name: String,
+    actual val version: String,
+    actual val packages: MutableList<Package>
```
marandaneto, line 6:
```suggestion
    actual val packages: List<Package> = mutableListOf()
```

Can we do this instead? or even make it nullable

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/protocol/SdkVersion.kt
buenaflor: This works but requires casting like `toCocoaSdkVersion() as Map<Any?, *>` when setting the sdk version on the cocoa sdk.

buildSrc/src/main/java/Config.kt
buenaflor: Another alternative could be `sentry.java.android.kotlin-multiplatform` or `sentry.cocoa.kotlin-multiplatform` depending on which platform the user uses. I guess this is more fitting since we are using multiple SDKs and the docs specify it using unity as an example

sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryAndroidOptionsExtensions.kt
buenaflor: `it.sdkVersion` can be null but `toJvmSdkVersion()` cannot

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryOptions.kt
```diff
@@ -56,4 +57,7 @@ open class SentryOptions {
 
     /** Hook that is triggered before a breadcrumb is sent to Sentry */
     var beforeBreadcrumb: ((Breadcrumb) -> Breadcrumb)? = null
+
+    /** Information about the Sentry SDK that generated this event. */
+    var sdk: SdkVersion = SdkVersion()
```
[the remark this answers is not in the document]
buenaflor: are there benefits to initializing it in a ctor? kotlin also keeps prompting to join declaration and assignment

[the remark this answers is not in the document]
marandaneto: its a semantical thing here, `SdkVersion` should have `name` and `version` as required, if you are able to provide both during declaration and assignment, fine, its not the case right now.

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/protocol/SdkVersion.kt
marandaneto: I believe the data class should have the correct type, when calling `toCocoaSdkVersion` you can do the casting.

buildSrc/src/main/java/Config.kt
marandaneto: I guess I mistakenly got confused with the dot and dash.
Maybe `sentry.kotlin.kmp?`

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtensions.kt
```diff
@@ -28,11 +30,29 @@ internal fun CocoaSentryOptions.applyCocoaBaseOptions(options: SentryOptions) {
...
+            }
+        }
+        val sdk = event?.sdk?.toMutableMap()
+        sdk?.set(
+            "packages",
+            options.sdk.packages?.map {
+                mapOf("name" to it.name, "version" to it.version)
+            }
+        )
+        event?.sdk = sdk
```
buenaflor, line 49: there's no way to attach packages in the cocoa SDK via PrivateSentrySDKOnly, so it is currently done in the beforeSend hook. Although I'm not sure if this is a reliable way to attach them, any thoughts on this?

brustolin: I think `beforeSend` is the only place that is possible to do this now. But what happens if the user set a `beforeSend` callback?

buenaflor: Right now the kmp sdk doesn't have support for a public facing `beforeSend` hook but it's possible to add these packages regardless of what the user sets in their `beforeSend`

build.gradle.kts
marandaneto: comment is resolved but Indentation is still off

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtensions.kt
marandaneto, line 49: Can we do that during the SDK init instead using the `PrivateSentrySDKOnly`?
Packages need to be added in beforeSend, AFAIK there's no other way yet https://github.com/getsentry/sentry-dart/blob/11fb40841ad3d6fc46e02eb8adeb74d5b6932582/flutter/ios/Classes/SentryFlutterPluginApple.swift#L233-L250

marandaneto: Packages need to be added in beforeSend, AFAIK there's no other way yet https://github.com/getsentry/sentry-dart/blob/11fb40841ad3d6fc46e02eb8adeb74d5b6932582/flutter/ios/Classes/SentryFlutterPluginApple.swift#L233-L250
https://github.com/getsentry/sentry-cocoa/issues/2270

build.gradle.kts
buenaflor: hmm it looks fine on my end or do you mean that it needs line breaks due to the length?

buenaflor:
@marandaneto I was thinking about adding event.origin and event.environmen tags as well like in flutter and react-native:

https://github.com/getsentry/sentry-dart/blob/11fb40841ad3d6fc46e02eb8adeb74d5b6932582/flutter/ios/Classes/SentryFlutterPluginApple.swift#L393

would that make sense here as well?

build.gradle.kts
buenaflor: https://github.com/getsentry/sentry-kotlin-multiplatform/blob/df6bce7dfe31a168e682abeda3e927ed7ac12ca0/build.gradle.kts#L12

marandaneto:
> @marandaneto I was thinking about adding event.origin and event.environment tags as well like in flutter and react-native:
> 
> [getsentry/sentry-dart@`11fb408`/flutter/ios/Classes/SentryFlutterPluginApple.swift#L393](https://github.com/getsentry/sentry-dart/blob/11fb40841ad3d6fc46e02eb8adeb74d5b6932582/flutter/ios/Classes/SentryFlutterPluginApple.swift#L393)
> 
> would that make sense here as well?

Nope because of https://github.com/getsentry/team-mobile/issues/60
Also, I'd focus on shipping the current version.

build.gradle.kts
marandaneto: I guess its the code diff on github, here its clearly off but opening the commit, it is right indeed.

marandaneto:
@buenaflor after double-checking this on Flutter and RN.
See my PR https://github.com/getsentry/sentry-dart/pull/1320
The Native SDKs should just have `.kmp` at the end, and not replace the whole thing.
Can we fix this?

PR #53, opened 2023-03-10
feat: view hierarchy

buenaflor:
Add view hierarchy option for Android and iOS. This requires bumping the Cocoa SDK version from `8.2.0` -> `8.4.0`

Testing:
 - iOS: https://sentry-sdks.sentry.io/issues/3963269927/?project=5903800&referrer=project-issue-stream
 - Android: https://sentry-sdks.sentry.io/issues/4046046165/?project=5903800&query=is%3Aunresolved&referrer=issue-stream&stream_index=0


sentry-kotlin-multiplatform/build.gradle.kts
```diff
@@ -82,7 +82,11 @@ kotlin {
             summary = "Official Sentry SDK Kotlin Multiplatform"
             homepage = "https://github.com/getsentry/sentry-kotlin-multiplatform"
 
-            pod(Config.Libs.sentryCocoa, "~> ${Config.Libs.sentryCocoaVersion}")
+            pod(Config.Libs.sentryCocoa) {
+                source = git("https://github.com/getsentry/sentry-cocoa.git") {
+                    tag = Config.Libs.sentryCocoaVersion
+                }
+            }
```
marandaneto, line 89: Why are we changing the source to `GH`?

buenaflor: Installing through cocoapods directly did not resolve for 8.3.0 (not sure how in sync they are with gh releases) and I was trying to keep it close to [sentry docs](https://docs.sentry.io/clients/cocoa/) since they also use gh

marandaneto: it should be something else, Cocoapods and GH should have the same version, for every release.
https://cocoapods.org/pods/Sentry#changelog
Let's focus on releasing the current version instead of fixing this now.

buenaflor: turned out to be a local gradle cache error so using cocoapods still works as expected

buenaflor:
What's missing is updating the sentry cocoa version to 8.3.0 in the readme but will wait for #56 first

marandaneto (approved):

PR #55, opened 2023-03-13
use craft 1.2.1

buenaflor:
Craft 1.2.1 will include the fix.

Can be merged as soon as 1.2.1 is released


marandaneto (approved):
the `publish` repo already picks up the latest version of craft, so it's fine to change this but you can also change it after being sure it works, otherwise you have to patch this for every craft patch.

PR #56, opened 2023-03-14
chore: update readme versions

buenaflor:
- change `mavenLocal` to `mavenCentral`
- added maven central badge 
- new gradle task that updates the versions used in the README which runs during the `format` gh actions workflow

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -29,6 +29,9 @@ jobs:
       - name: Make format
         run: make format
 
+      - name: Update README
+        run: make update-readme
+ 
       # actions/checkout fetches only a single commit in a detached HEAD state. Therefore
       # we need to pass the current branch, otherwise we can't commit the changes.
       # GITHUB_HEAD_REF is the name of the head branch. GitHub Actions only sets this for PRs.
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -25,7 +25,6 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
           restore-keys: |
             ${{ runner.os }}-gradle-
-            
       - name: Make format
         run: make format
 
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -25,6 +25,9 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
           restore-keys: |
             ${{ runner.os }}-gradle-
+            
+      - name: Make format
+        run: make format
 
       # actions/checkout fetches only a single commit in a detached HEAD state. Therefore
       # we need to pass the current branch, otherwise we can't commit the changes.
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -26,9 +26,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
 
-      - name: Make format
-        run: make format
-
       - name: Update README
         run: make update-readme
  
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -26,9 +26,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
 
-      - name: Update README
-        run: make update-readme
- 
       # actions/checkout fetches only a single commit in a detached HEAD state. Therefore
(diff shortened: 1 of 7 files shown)


sentry-kotlin-multiplatform/sentry_kotlin_multiplatform.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
     spec.name                     = 'sentry_kotlin_multiplatform'
-    spec.version                  = '0.0.1'
+    spec.version                  = '0.0.1-alpha.2'
```
buenaflor, line 3: I'll have to figure out in another PR to update this automatically as it changes to the project's version during gradle build.

build.gradle.kts
```diff
@@ -50,6 +50,26 @@ subprojects {
...
+    val kmpRegex = Regex("io\\.sentry:sentry-kotlin-multiplatform:[^\"]*")
+    val cocoaRegex = Regex("pod\\(\"Sentry\", \"~>[^\"]*\"\\)")
+
+    doLast {
+        try {
+            val readmeFile = File(filename)
+
+            val newContents = readmeFile.readText()
+                .replace(kmpRegex, "io.sentry:sentry-kotlin-multiplatform:${project.version}")
+                .replace(cocoaRegex, "pod(\"Sentry\", \"~> ${Config.Libs.sentryCocoaVersion}\")")
```
marandaneto, line 64: The project version isn't necessarily published yet.
I believe that using the maven badges with the latest version (similar to sentry-java) and using a placeholder in the versioning is more effective.

sentry-kotlin-multiplatform/sentry_kotlin_multiplatform.podspec
marandaneto: For Flutter is always `0.0.1`, so it'd be fine to keep it as it is.

README.md
```diff
@@ -24,36 +28,29 @@ This SDK is a wrapper around different platforms such as JVM, Android, iOS, macO
...
+    api("io.sentry:sentry-kotlin-multiplatform:0.0.1-alpha.2")
   }
 }
 ```
 
 ### Cocoa
 
 If you are targeting Apple platforms (iOS, macOS, watchOS, tvOS), then you need to use CocoaPods to include [Sentry Cocoa](https://github.com/getsentry/sentry-cocoa) into this SDK.
-One way to achieve this is to include the Sentry Cocoa SDK via the Kotlin CocoaPods extension. Be aware that your Sentry Cocoa version has to match the SDK's version. Currently the supported version is `~> 7.21.0` 
+One way to achieve this is to include the Sentry Cocoa SDK via the Kotlin CocoaPods extension. Be aware that your Sentry Cocoa version has to match the SDK's version. Currently the supported version is `pod("Sentry", "~> 8.2.0")`
```
marandaneto, line 53: Is there a way to do this automatically though? so People install the KMP SDK and that pulls the compatible version of the Cocoa SDK?

buenaflor: Not yet unfortunately. Users need to pull in the cocoa sdk dep themselves as KMP afaik only includes the headers in their apple packages so the actual implementation of these headers has to be provided from somewhere else. 

that also means that users aren't necessarily tied to using cocoapods. but I haven't tested yet if SPM or Carthage works (theoretically should though).

build.gradle.kts
buenaflor: Maybe this could only be run on release branches? This would be useful for managing the required sentry cocoa version otherwise we'd need to manage that manually in the README

buenaflor: Sounds good - I'd like to at least keep the required sentry cocoa sdk version documented for the user

buenaflor: As sentry-docs needs to be in sync with cocoa version as well, we should do it manually for now https://github.com/getsentry/sentry-docs/pull/6452#discussion_r1136921094

.github/workflows/format.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
           restore-keys: |
             ${{ runner.os }}-gradle-
-
+
```
marandaneto, line 28:
```suggestion

```

sentry-kotlin-multiplatform/sentry_kotlin_multiplatform.podspec
marandaneto, line 3:
```suggestion
    spec.version                  = '0.0.1'
```

PR #61, opened 2023-03-24
fix: Android init crash on API level 23 and below

### LINKED ISSUE #60: Crashes during init on Android API 23 and below
### Environment

Android API level 23 and below

### Steps to Reproduce

1. Use an Android simulator with API level 23 or below in the sample project
### Log
https://sentry-sdks.sentry.io/issues/4026916773/?project=5903800&query=is%3Aunresolved&referrer=issue-stream

```
AbstractMethodError
abstract method "boolean io.sentry.util.thread.IMainThreadChecker.isMainThread()"
```

buenaflor:
Fixes #60 

Reason is that on API 23 and below `SentryAndroid` depends on some `Sentry` functions internally so that's why we cannot exclude it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- fix: crash on Android API levels 23 and below ([#61](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/61))
+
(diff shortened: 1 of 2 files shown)


marandaneto (approved):

marandaneto:
@buenaflor Add to the changelog

PR #62, opened 2023-03-29
feat: improve Objc/Swift experience with HiddenFromObjc

buenaflor:
Add `@HiddenFromObjc` annotation to hide `init` with context from Cocoa targets.
This is very useful if we want to have certain things hidden to make the Swift experience better in general

https://kotlinlang.org/docs/native-objc-interop.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 # Changelog
 
-## 0.0.3
+## Unreleased
 
 ### Features
 
 - Improve Objc/Swift experience with @HiddenFromObjc ([#62](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/62))
+
+## 0.0.3
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.0.3
 
+### Features
+
+- Improve Objc/Swift experience with @HiddenFromObjc ([#62](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/62))
+
 ### Fixes
 
 - fix: crash on Android API levels 23 and below ([#61](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/61))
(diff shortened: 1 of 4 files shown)


sentry-samples/kmp-app/iosApp/iosApp/iOSApp.swift
```diff
@@ -3,15 +3,15 @@ import shared
 
 @main
 struct iOSApp: App {
-    let sentry = Sentry()
-
+    let sentry = Sentry.shared
```
buenaflor, line 6: Singletons mapped via Kotlin should be accessed via `.shared` or `.companion` or `Companion.shared`: https://kotlinlang.org/docs/native-objc-interop.html#kotlin-singletons

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 0.0.3
 
+### Features
+
+- Improve Objc/Swift experience with @HiddenFromObjc ([#62](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/62))
```
marandaneto, line 7: That should be under the ## Unreleased entry

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryKMP.kt
```diff
@@ -19,6 +21,8 @@ object Sentry {
      * @param context: The context (used for retrieving Android Context)
      * @param configuration Options configuration handler.
      */
+    @OptIn(ExperimentalObjCRefinement::class)
+    @HiddenFromObjC
```
marandaneto, line 25: M: Does this require a min. version of KMP that would not work in older versions?
If yes, maybe we can consider just adding a comment that this is not for Apple folks.

marandaneto (approved):

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryKMP.kt
buenaflor: The min. version for using our KMP sdk is `1.8.0` so nothing would change for current users, I should probably add a badge or something similar to our readme

PR #63, opened 2023-03-29
Ref: use explicit api & consistency

buenaflor:
Add explicit api to make sure nothing is exposed unnecessarily.

Also improve consistency of the code and remove any unnecessary interfaces that is not used at all.


marandaneto (approved):
Good idea with the `explicitApi`

Issue #65, opened 2023-03-31
Support configuration of fingerprint for Events


Legion2:
To group issues and messages in the sentry ui, normally the stack trace is used, but we want to customize the grouping for specific issues. The Android and IOS sdk allow to set a fingerprint for issues and messages, which used to group the issues in sentry.

It should be possible to set the fingerprint via the multiplatform sdk.

Issue #47, opened 2023-02-21
Support configuration of beforeSend hook


Legion2:
The sentry SDK for android and ios have both the beforeSend filter which allows to filter and manipulate events before they are send. Currently it is not possible to set the beforeSend filter in kotlin multiplatform.

buenaflor:
Thanks for the suggestion, it's definitely on the list. Right now, we're focused on getting the first release out the door, so we may not be able to work on it immediately. But we'll definitely keep it in mind for future updates.

buenaflor:
We've shipped this in `0.1.0`: https://github.com/getsentry/sentry-kotlin-multiplatform/releases/tag/0.1.0

buenaflor:
Please beware that for `0.1.0` there is currently an issue where the SDK will drop events if `beforeSend` is not configured.
So to be on the safe side make sure to include `beforeSend` during `Sentry.init`. See #79 

We will put out the fix asap

buenaflor:
Now fixed on `0.1.1`: https://github.com/getsentry/sentry-kotlin-multiplatform/releases/tag/0.1.1

PR #70, opened 2023-04-06
feat: beforeSend / fingerprinting

buenaflor:
Ref: #47 & #65 

This adds a `beforeSend` hook to manipulate event - at the same time it also adds the capability to add fingerprints.


gradle.properties
```diff
@@ -1,5 +1,4 @@
 # Kotlin
-kotlin.version=1.8.0
```
buenaflor, line 2: after some investigation this is never used. we define our Kotlin version in `buildSrc/src/config` so it's fine to delete it.

sentry-kotlin-multiplatform/src/commonJvmMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryOptionsExtensions.kt
```diff
@@ -39,4 +43,9 @@ internal fun JvmSentryOptions.applyJvmBaseOptions(options: SentryOptions) {
     this.setBeforeBreadcrumb { jvmBreadcrumb, _ ->
         options.beforeBreadcrumb?.invoke(jvmBreadcrumb.toKmpBreadcrumb())?.toJvmBreadcrumb()
     }
+    this.setBeforeSend { jvmSentryEvent, hint ->
+        options.beforeSend?.invoke(SentryEvent(jvmSentryEvent))?.let {
+            jvmSentryEvent.applyKmpEvent(it)
```
buenaflor, line 48: in all cases we don't want to replace the event but only modify / apply changes to it

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/protocol/User.kt
```diff
@@ -2,13 +2,13 @@ package io.sentry.kotlin.multiplatform.protocol
 
 public data class User(
     /** The user's email */
-    var email: String = "",
+    var email: String? = null,
```
buenaflor, line 5: default values of users should be null, not empty strings

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/UserExtensions.kt
```diff
@@ -5,17 +5,16 @@ import io.sentry.kotlin.multiplatform.protocol.User
 
 internal fun User.toCocoaUser() = CocoaUser().apply {
     val scope = this@toCocoaUser
-    userId = scope.id
-    username = scope.username
-    email = scope.email
-    ipAddress = scope.ipAddress
+    this.userId = scope.id
```
marandaneto, line 8: I don't think you need `this` when using `apply`, or is this because of the extension method?

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryBaseEvent.kt
```diff
@@ -0,0 +1,55 @@
...
+        }
+        this.mutableTags!![key] = value
+    }
+
+    public fun addBreadcrumb(breadcrumb: Breadcrumb) {
+        if (mutableBreadcrumbs == null) {
+            this.mutableBreadcrumbs = mutableListOf()
+        }
+        this.mutableBreadcrumbs?.add(breadcrumb)
+    }
```
marandaneto, line 50: I rather prefer using `?` instead of `!!` otherwise you may crash the app.

marandaneto: This code has concurrency issues if 2 threads are calling at the same time.
A simple way of solving this is always creating the list during declaration using a thread-safe type.

marandaneto:
I'd like to see some unit tests with all the other changes.

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/SentryEvent.kt
```diff
@@ -0,0 +1,51 @@
...
+    public actual var exceptions: List<SentryException>? = null
+    public override var release: String? = null
+    public override var environment: String? = null
+    public override var platform: String? = null
+    public override var user: User? = null
+    public override var serverName: String? = null
+    public override var dist: String? = null
+
+    public constructor(cocoaSentryEvent: CocoaSentryEvent?) : this() {
+        this.eventId = SentryId(cocoaSentryEvent?.eventId.toString())
```
marandaneto, line 28: This probably will throw an exception if null?

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryBaseEvent.kt
buenaflor: Good point, didn't think of possible concurrency issues, will adjust it.

buenaflor: Okay so one problem is that something like `ConcurrentHashMap` comes from the Java lib but I don't have access to that here since this is in the shared `commonMain` module.

Alternatives:
 - https://github.com/touchlab/Stately use something like this, concurrent collections specifically designed for KMP

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/UserExtensions.kt
buenaflor: We don't need it but since other code parts were using it I wanted to keep things consistent

Unless you think it's better code practice to omit `this` when possible?

marandaneto: One of the goals of scoped functions is to reduce boilerplate.
https://kotlinlang.org/docs/scope-functions.html
If you use eg `user.userId`, `user.email` the boiler plate is still there, from `user` to `this` is the same thing IMO.
You need to use `this` in case there's a shadowing issue tho, if it's not the case, I'd prefer to avoid `this`.

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryBaseEvent.kt
marandaneto: What you can do for now is, add a comment this is not thread-safe for every public list/map.
Raise an issue for making those things thread-safe, but let's not block this for now.

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/protocol/User.kt
buenaflor: It's for convenience in initializing Users since we're also defining the constructor with this in this data class

buenaflor:
@marandaneto bump, you think it's good2go?

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryBaseEvent.kt
```diff
@@ -0,0 +1,52 @@
...
+    public open var serverName: String? = null
+    public open var dist: String? = null
+
+    /** This is not thread-safe */
+    public open val contexts: Map<String, Any>? get() = mutableContexts
+    internal var mutableContexts: Map<String, Any>? = null
+
+    /** This is not thread-safe */
+    public open val breadcrumbs: List<Breadcrumb>? get() = mutableBreadcrumbs
+    internal var mutableBreadcrumbs: MutableList<Breadcrumb>? = null
```
marandaneto, line 23: Why do we need this instead of doing it directly?

buenaflor: You're right, we don't need this for breadcrumbs, tags, fingerprints and exceptions.

However, I'd like to keep it for context to keep context immutable as I'll introduce a separate `Context` class later on to modify context much more easily with less code duplication.

marandaneto: It's not immutable, the getter just returns the `mutableBreadcrumbs`, the only difference is the interface, or am I missing something? If you want something immutable, you need an immutable type such as.
`val immutableList: List<String> = mutableList.toList()`

buenaflor: Yes it's not truly immutable but it prevents the user from modifying it by exposing only a `List` so to be exact we are exposing a read-only List, not an immutable one.

However, we don't need that anymore as we can also do this:

```
public var contexts: Map<String, Any>? = mapOf()
    internal set
```

Since we only care about being able to set the context internally but not publicly.

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/SentryBaseEvent.kt
```diff
@@ -0,0 +1,44 @@
...
+    public open var user: User? = null
+    public open var serverName: String? = null
+    public open var dist: String? = null
+
+    /** This is not thread-safe */
+    public var contexts: Map<String, Any>? = mapOf()
+        internal set
+
+    /** This is not thread-safe */
+    public open var breadcrumbs: MutableList<Breadcrumb>? = mutableListOf()
```
marandaneto, line 22: L: All those lists, maps, etc don't need to be nullable if you are initing during declaration with an empty value.
It's fine as it is, somebody can still set it to `null`, but it'd be natural to just do `breadcrumbs.clear()`.

Issue #17, opened 2022-07-19
Tests capturing messages and exceptions on iOS are not being sent to the sentry project


buenaflor:
### Problem Statement
Capturing messages and exceptions on iOS tests doesn't send them to the sentry project. 
- Due to threading, the test finishes before this async block finishes: https://github.com/getsentry/sentry-cocoa/blob/c51bd77615c8685569ad3da7bad7a001479ce110/Sources/Sentry/SentryHttpTransport.m#L92 (easily fixed)
- SSL errors occur when trying to send an envelope

marandaneto:
@buenaflor does the iOS SDK has a `flush` method that awaits the async dispatch to be finished?

marandaneto:
@buenaflor can you link the SSL issue? that is strange, we didn't run into this issue anywhere else.

buenaflor:
@marandaneto 

I have a fix ready for the async part but the largest problem is the ssl error.
Here is a snippet of the error:

```
Request failed: Error Domain=NSURLErrorDomain 
Code=-1200 
"An SSL error has occurred and a secure connection to the server cannot be made." 
UserInfo={
    NSErrorFailingURLStringKey=https://o447951.ingest.sentry.io/api/5903800/envelope/, 
    NSLocalizedRecoverySuggestion=Would you like to connect to the server anyway?, 
    _kCFStreamErrorDomainKey=3, 
    _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <C6DB900D-8382-4009-8606-7860E11552FA>.<2>, 
    _NSURLErrorRelatedURLSessionTaskErrorKey=(
        "LocalDataTask <C6DB900D-8382-4009-8606-7860E11552FA>.<2>"
    ), 
    NSLocalizedDescription=An SSL error has occurred and a secure connection to the server cannot be made., 
    NSErrorFailingURLKey=https://o447951.ingest.sentry.io/api/5903800/envelope/, 
    NSUnderlyingError=0x6000020d50e0 {
        Error Domain=kCFErrorDomainCFNetwork 
        Code=-1200 
        "(null)" 
        UserInfo={
            _kCFStreamPropertySSLClientCertificateState=0, 
            _kCFNetworkCFStreamSSLErrorOriginalValue=-9836, 
            _kCFStreamErrorDomainKey=3, 
            _kCFStreamErrorCodeKey=-9836, 
[...]

PR #78, opened 2023-04-24
test: add sentry integration & e2e tests

buenaflor:
This introduces Sentry integration and e2e tests. e2e tests are disabled for Apple targets temporarily as they have some issues regarding SSL that are not fixed very easily. (#17)


sentry-kotlin-multiplatform/src/androidUnitTest/kotlin/io/sentry/kotlin/multiplatform/BaseSentryTest.kt
```diff
@@ -0,0 +1,17 @@
+package io.sentry.kotlin.multiplatform
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+@Config(sdk = [28])
```
marandaneto, line 9: Is there a reason to use 28 specifically?

sentry-kotlin-multiplatform/src/commonTest/kotlin/io/sentry/kotlin/multiplatform/utils/Constants.kt
```diff
@@ -0,0 +1,5 @@
+package io.sentry.kotlin.multiplatform.utils
+
+const val fakeDsn =
+    "https://00f281ded2844eda83a8a413b080dbb9@o447951.ingest.sentry.io/5903800"
```
marandaneto, line 4:
```suggestion
const val fakeDsn =
    "https://abc@def.ingest.sentry.io/1234567"
```

marandaneto: Its easier to understand that its fake.

sentry-kotlin-multiplatform/src/commonTest/kotlin/io/sentry/kotlin/multiplatform/SentryIntegrationTest.kt
```diff
@@ -0,0 +1,266 @@
...
+                try {
+                    assertNotNull(contexts)
+                    assertEquals(mapOf("value" to stringValue), contexts[stringKey])
+                    assertEquals(mapOf("value" to booleanValue), contexts[booleanKey])
+                    assertEquals(mapOf("value" to numberValue), contexts[numberKey])
+                    assertEquals(mapOf("value" to collectionValue), contexts[collectionKey])
+                    deferred.complete(Unit)
+                } catch (e: Throwable) {
+                    deferred.completeExceptionally(e)
+                }
```
marandaneto, line 251: We can assert after, no need of try-catch, otherwise, if the assert fails, the test fails anyway, or is there any other reason?

sentry-kotlin-multiplatform/src/androidUnitTest/kotlin/io/sentry/kotlin/multiplatform/BaseSentryTest.kt
marandaneto: It really depends if you are using specific features that depend on OS specific, otherwise, it's not important.

sentry-kotlin-multiplatform/src/androidUnitTest/kotlin/io/sentry/kotlin/multiplatform/BaseSentryTest.kt
```diff
@@ -0,0 +1,15 @@
+package io.sentry.kotlin.multiplatform
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+actual abstract class BaseSentryTest {
+    actual val platform: String = "Android"
+    actual val authToken: String? = System.getenv("SENTRY_AUTH_TOKEN")
```
marandaneto, line 10: IIRC `System.getenv` returns empty if no env. var. found, maybe we need to check that, or the usage of `authToken` should check empty/null.
Right now it does only `assertNotNull(authToken)` but empty is valid.

sentry-kotlin-multiplatform/src/commonTest/kotlin/io/sentry/kotlin/multiplatform/SentryE2ETest.kt
```diff
@@ -0,0 +1,130 @@
...
+import kotlin.time.Duration.Companion.seconds
+
+@Serializable
+private data class SentryEventSerializable(
+    val id: String? = null,
+    val project: Long? = null,
+    val release: String? = null,
+    val platform: String? = null,
+    val message: String? = "",
+    val tags: List<TagSerializable> = listOf(),
```
marandaneto, line 31: M: Should just be a Map then? no need having a `TagSerializable`

sentry-kotlin-multiplatform/src/commonTest/kotlin/io/sentry/kotlin/multiplatform/SentryE2ETest.kt
```diff
@@ -0,0 +1,130 @@
...
+            options.beforeSend = { event ->
+                sentEvent = event
+                event
+            }
+        }
+    }
+
+    private suspend fun fetchEvent(eventId: String): String {
+        val url =
+            "https://sentry.io/api/0/projects/sentry-sdks/sentry-kotlin-multiplatform/events/$eventId/"
```
marandaneto, line 63: M: Lets do similar to Dart, parametrize the URL.
https://github.com/getsentry/sentry-dart/blob/99490581cac83f47aff9c6d3186d401d3dbc55db/e2e_test/bin/e2e_test.dart#L99
Put the org and project closer to the DSN, so in case we change it, we know that we have to change the project and org too instead of only DSN.

marandaneto (approved):
Left a few comments otherwise LGTM

PR #85, opened 2023-05-05
chore: add api checks with binary compatibility plugin

buenaflor:
ensures that our API won't introduce breaking changes unnoticed.

Unfortunately native targets are not supported yet but it's a good start to have checks at least for JVM and Android

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 ### Improvements
 
-- chore: add api checks with binary compatibility plugin ([#85](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/85))
 - ref: improve samples & add SPM docs ([#82](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/82))
 
 ## 0.1.1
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Improvements
 
+- chore: add api checks with binary compatibility plugin ([#85](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/85))
```
marandaneto, line 7: CHANGELOG is meant for user-facing changes, no need to put those changes in the changelog.

Makefile
```diff
@@ -21,6 +25,7 @@ format:
 
 # build and run tests
 compile:
+	make checkApi
```
marandaneto, line 28: I think you don't need to call `make` but rather the function directly, maybe there's a way how to do that, if you find out how, you can remove the `make` from `buildAppleSamples` as well.
Also, when adding new functions, add it to the `.PHONY` line.

marandaneto (approved):

PR #86, opened 2023-05-08
chore: add detekt & dokka

buenaflor:
- Detekt for enforcing some doc rules and general code smells
 - Dokka will be published to gh-pages 

I've also updated the docs and naming of `Scope` to make it clear what it does.


detekt.yml
```diff
@@ -0,0 +1,15 @@
+comments:
+  UndocumentedPublicClass:
+    active: true
+    includes: "**/commonMain/**"
+  UndocumentedPublicProperty:
+    active: true
+    includes: "**/commonMain/**"
```
buenaflor, line 7: I think it makes sense to force documentation of public facing components at least in commonMain since this will be what users will work with

detekt.yml
```diff
@@ -0,0 +1,15 @@
...
+style:
+  UnnecessaryAbstractClass:
+    active: false
```
buenaflor, line 10: I've disabled this since it behaves weird with Kotlin Multiplatform and doesn't detect expect/actual usages

detekt.yml
```diff
@@ -0,0 +1,15 @@
...
+complexity:
+  TooManyFunctions:
+    thresholdInObjects: 21
+    thresholdInClasses: 21
```
buenaflor, line 14: Not sure if 21 is a reasonable default, any thoughts on this?

sentry-kotlin-multiplatform/api/jvm/sentry-kotlin-multiplatform.api
```diff
@@ -42,34 +42,7 @@ public final class io/sentry/kotlin/multiplatform/HttpStatusCodeRange {
...
-	public fun setContext (Ljava/lang/String;Z)V
-	public fun setContext (Ljava/lang/String;[Ljava/lang/Object;)V
-	public fun setExtra (Ljava/lang/String;Ljava/lang/String;)V
-	public fun setLevel (Lio/sentry/kotlin/multiplatform/SentryLevel;)V
-	public fun setTag (Ljava/lang/String;Ljava/lang/String;)V
-	public fun setUser (Lio/sentry/kotlin/multiplatform/protocol/User;)V
-}
-
-public abstract interface class io/sentry/kotlin/multiplatform/ScopeProvider {
+public abstract interface class io/sentry/kotlin/multiplatform/Scope {
```
buenaflor, line 45: this won't break compatibility since Users wouldn't have created a Scope themselves

.github/workflows/build.yml
```diff
@@ -55,3 +55,27 @@ jobs:
           if-no-files-found: error
           path: |
             ./*/build/distributions/*.zip
+
+  generate-dokka:
```
marandaneto, line 59: Lets split this up and only generate during a `release`, see https://github.com/getsentry/sentry-java/blob/main/.github/workflows/generate-javadocs.yml

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Improvements
 
 - ref: improve samples & add SPM docs ([#82](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/82))
+- ref: add documentation, detekt and dokka ([#86](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/86))
```
marandaneto, line 8: @buenaflor changelog is meant for user-facing changes.

detekt.yml
marandaneto: Detekt already has sensible defaults https://github.com/getsentry/sentry-java/blob/ba90bd9ebf765bac954ce09e164152229da553ef/build.gradle.kts#L102
Any reason why we are being explicit about magic numbers?

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/SentryUnhandledExceptions.kt
```diff
@@ -75,7 +75,7 @@ internal fun Throwable.asSentryEnvelope(): SentryEnvelope {
         hub.getClient()?.prepareEvent(event, hub.scope, alwaysAttachStacktrace = false, isCrashEvent = true)
     } ?: event
     val item = SentryEnvelopeItem(preparedEvent)
-    // TODO: pass traceState when enabling performance monitoring for KMP SDK
+    // TODOs: pass traceState when enabling performance monitoring for KMP SDK
```
marandaneto, line 78:
```suggestion
    // TODO: pass traceState when enabling performance monitoring for KMP SDK
```

sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/Attachment.kt
```diff
@@ -1,5 +1,7 @@
 package io.sentry.kotlin.multiplatform
 
+@Suppress("UnusedPrivateMember")
```
marandaneto, line 3: Usually those annotations just pollute the API, would be possible to suppress via a generated file? for example, https://github.com/getsentry/sentry-dart/blob/main/flutter/config/detekt-bl.xml

CHANGELOG.md
buenaflor: Ah, I thought adding API documentation is part of user facing

detekt.yml
buenaflor: They probably updated it but the default now includes explicit magic numbers: https://github.com/detekt/detekt/blob/292432870e9382b7b0aad1fb7c17964027d69682/detekt-core/src/main/resources/default-detekt-config.yml#L606

As for TooManyFunctions: the default is 11 so I guess it would be fine to exclude files like SDK entry points?

sentry-kotlin-multiplatform/src/commonAppleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/SentryUnhandledExceptions.kt
buenaflor: Detekt will flag TODO so one option would be to disable it, the other is to Suppress the line but that doesn't work since ktlint will format it in such way that the `@Suppress` will be in another line.

That aside though, I don't think we need this TODO, this specific envelope is only created during a crash and the purpose is only to send the crash event and won't be used for anything else.

CHANGELOG.md
marandaneto: Yeah, it could be, detekt and dokka not really, but I agree, your call here.

.github/workflows/generate-dokka.yml
```diff
@@ -0,0 +1,52 @@
...
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+jobs:
+  generate-dokka:
+    runs-on: macos-latest
```
marandaneto, line 15: Usually, linux machines are more available unless macos is needed.

config/detekt/detekt.yml
```diff
@@ -0,0 +1,25 @@
...
+complexity:
+  TooManyFunctions:
+    excludes: [
+      "**/SentryKMP.kt",
+      "**/SentryBridge.kt",
+      "**/CocoaScopeProvider.kt",
+      "**/Scope.kt",
+      "**/JvmScopeProvider.kt",
+      "**/Breadcrumb.kt",
+    ]
```
marandaneto, line 25:
```suggestion
    ]

```

marandaneto (approved):

Issue #88, opened 2023-05-12
Filename guidelines to expect/actual and platform specific implementations


buenaflor:
As our codebase grows it's important to keep it organized in a way that makes searching for specific implementations easy. 

Example: we specify an `expect class SentryEvent` in our `commonMain` which is called `SentryEvent.kt`.
Now on `commonAppleMain` and `commonJvmMain` the `actual` implementations are also called `SentryEvent.kt` so we have 3 files with the same name.

How Google for example handles it: suffix the filename with the platform -> `SentryEvent.jvm.kt` or `SentryEvent.apple.kt`.

[See here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:collection/collection/src/jvmMain/kotlin/androidx/collection/ArraySet.jvm.kt) for an example.

This is not publicly endorsed by Google but could be a good starting point on how to organize this.

This could also not be only restricted to expect actuals but in general for any File name that will exist on multiple platforms with the same name. So this could also be helpful for our extensions: `SentryOptionsExtensions.jvm.kt` or `BreadcrumbExtensions.jvm.kt`

Thoughts?

buenaflor:
@marandaneto

marandaneto:
@buenaflor makes sense, if we are adding more and more actual classes, sounds like a good improvement.

buenaflor:
Already implemented and is now the standard going forward

PR #100, opened 2023-07-04
chore: change filenames of multiplatform files to have a suffix of the respective platforms

buenaflor:
Related issue: #88 

Filenames can get confusing in Kotlin Multiplatform if you have multiple files with the same name in different platforms.

This PR introduces filename changes that adds the platform suffix to files that are available in multiple platforms (typically those files exist because of `expect / actual`).

Google also adheres to this guideline, example: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:collection/collection/src/jvmMain/kotlin/androidx/collection/ArraySet.jvm.kt

#skip-changelog


buenaflor:
Closing due to issues with secrets not working through forks

Issue #99, opened 2023-06-29
Breadcrumbs Don't Seem to Work Right on iOS


alexwhb:
### Environment
iOS 

0.2.0 for the shared kmm code
and 8.4.0 for the iOS library using SPM 

### Steps to Reproduce
I wrote a custom log writer that adds bread crumbs to my sentry log every time something is globally logged in my app. 
This works great on Android and will add all the breadcrumbs to my error / crash log, but on iOS the only way I can seem to get a breadcrumb to log with my error / exception is to write it in the capture exception scope like this: 
```kotlin
Sentry.captureException(throwable) {
    it.addBreadcrumb(Breadcrumb.debug("TEST"))
}
```
### Expected Result
Expected that the last 100 breadcrumb before the exception will be reported since 100 is the default config value. Obviously assuming I logged that many things before the error / exception was logged. 

### Actual Result
I only see my "TEST" breadcrumb because that one is recorded in the scope. Again this is only a weird behavior on iOS. It works fine on Android. 

Here's my logger code that is written in my common kotlin module and works fine for Android, but not for iOS: 

```kotlin 
class SentryLogWriter(
    private val minSeverity: Severity,
    private val minCrashSeverity: Severity,
) : LogWriter() {
    init {
        check(minSeverity <= minCrashSeverity) {
            "minSeverity ($minSeverity) cannot be greater than minCrashSeverity ($minCrashSeverity)"
        }
    }

    override fun isLoggable(severity: Severity): Boolean = severity >= minSeverity

    override fun log(severity: Severity, message: String, tag: String, throwable: Throwable?) {
        if (throwable == null) {
            Sentry.addBreadcrumb(
                Breadcrumb(
                    severity.toSentryLevel(),
                    message = message,
                    category = tag
                )
            )
        }
        if (throwable != null && severity >= minCrashSeverity) {
            Sentry.captureException(throwable) {
                it.addBreadcrumb(Breadcrumb.debug("TEST this is a Test"))
            }
        }
    }
}

fun Severity.toSentryLevel(): SentryLevel {
    return when (this) {
        Severity.Debug -> SentryLevel.DEBUG
        Severity.Info -> SentryLevel.INFO
        Severity.Warn -> SentryLevel.WARNING
        Severity.Error -> SentryLevel.ERROR
        Severity.Assert -> SentryLevel.FATAL
        else -> throw IllegalArgumentException("Unsupported severity level")
    }
}
```

[...]

PR #101, opened 2023-07-06
fix: beforeBreadcrumb discarding breadcrumbs on Apple platforms

buenaflor:
Related: #99 

Breadcrumbs on Apple platforms are discarded if the hook `beforeBreadcrumb` is not configured.

Due to optional chaining it treats it the same as 

```
beforeBreadcrumb = {
  null
}
```


buenaflor:
Closing in favor of #101 due to issues with env secrets not working when workflow is triggered by forks (e2e tests are failing)

PR #105, opened 2023-07-07
fix: beforeBreadcrumb

buenaflor:
Related: https://github.com/getsentry/sentry-kotlin-multiplatform/issues/99

Breadcrumbs on Apple platforms are discarded if the hook beforeBreadcrumb is not configured.

Due to optional chaining it treats it the same as

```
beforeBreadcrumb = {
  null
}
```


romtsn (approved):
nice!

PR #124, opened 2023-10-25
Add sentry dependency bot

### LINKED ISSUE #123: Add dependency bot for bumping Cocoa and Java/Android SDKs
### Description

We can use the implementation from here: https://github.com/getsentry/sentry-react-native/actions/runs/6621794162/workflow

buenaflor:
Closes #123

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -1,0 +1,31 @@
+name: Update Dependencies
+
+on:
+  # Run every day.
+  schedule:
+    - cron: '0 3 * * *'
+  # And on on every PR merge so we get the updated dependencies ASAP, and to make sure the changelog doesn't conflict.
+  push:
+    branches:
+      - main
+
+jobs:
+  cli:
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v1
+    with:
+      path: sentry-cli.properties
+      name: CLI
+    secrets:
+      # If a custom token is used instead, a CI would be triggered on a created PR.
+      # api_token: ${{ secrets.CI_DEPLOY_KEY }}
+      api_token: ${{ github.token }}
+
+  android:
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v1
+    with:
+      path: scripts/update-cocoa.sh
+      name: Cocoa SDK
+    secrets:
+      # If a custom token is used instead, a CI would be triggered on a created PR.
+      # api_token: ${{ secrets.CI_DEPLOY_KEY }}
+      api_token: ${{ github.token }}
--- a/scripts/update-cocoa.sh
+++ b/scripts/update-cocoa.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+cd $(dirname "$0")/../
+file='buildSrc/src/main/java/Config.kt'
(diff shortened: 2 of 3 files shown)


buenaflor:
Example: https://github.com/getsentry/sentry-kotlin-multiplatform/pull/127

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  cli:
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v1
+    with:
+      path: sentry-cli.properties
+      name: CLI
+    secrets:
+      # If a custom token is used instead, a CI would be triggered on a created PR.
+      # api_token: ${{ secrets.CI_DEPLOY_KEY }}
+      api_token: ${{ github.token }}
```
krystofwoldrich, line 21: Maybe I missed it, but `sentry-cli` doesn't ship with `sentry-kmp`. 

The value from `sentry-cli.properties` is not used.

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,41 @@
...
+
+  cocoa:
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v1
+    with:
+      path: scripts/update-cocoa.sh
+      name: Cocoa SDK
+    secrets:
+      # If a custom token is used instead, a CI would be triggered on a created PR.
+      # api_token: ${{ secrets.CI_DEPLOY_KEY }}
```
krystofwoldrich, line 30: That's strange. I've only seen `api_token: ${{ secrets.CI_DEPLOY_KEY }}` used in `update-deps.yml`.

scripts/update-cocoa.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+cd $(dirname "$0")/../
+file='buildSrc/src/main/java/Config.kt'
```
krystofwoldrich, line 4: https://github.com/getsentry/sentry-kotlin-multiplatform/blob/ea9b2a45c737ddb46b1076114739a62405de7218/sentry-kotlin-multiplatform/sentry_kotlin_multiplatform.podspec#L15

Is the `podspec` updated from the `Config.kt`? Or do we need to update it too?

krystofwoldrich:
Nit: The files are missing a new empty line at the end.

.github/workflows/update-deps.yml
buenaflor: yup, that's an oversight, I somehow assumed the action itself needed the properties file

.github/workflows/update-deps.yml
buenaflor: Both work but yeah `secrets.CI_DEPLOY_KEY` should be more flexible

scripts/update-cocoa.sh
buenaflor: Good catch, we need to update it as well

krystofwoldrich:
@buenaflor Yeah, it only shows when the lines at the end are missing.

krystofwoldrich (approved):
LG 🚀

PR #171, opened 2023-12-20
fix: updater config v2

buenaflor:
#skip-changelog

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -8,23 +8,20 @@ on:
   push:
     branches:
       - main
-  pull_request:
 
 jobs:
   cocoa:
-    uses: getsentry/github-workflows/.github/workflows/updater.yml@v1
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v2
     with:
       path: scripts/update-cocoa.sh
       name: Cocoa SDK
     secrets:
-      # If a custom token is used instead, a CI would be triggered on a created PR.
-      api_token: ${{ secrets.CI_DEPLOY_KEY }}
+      api-token: ${{ secrets.CI_DEPLOY_KEY }}
 
   java:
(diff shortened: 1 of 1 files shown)


buenaflor:
Works on @v1

bruno-garcia (approved):

bruno-garcia:
@vaind seems like v2 breaks because of a token thing but v1 works. 

But repo has `CI_DEPLOY_KEY` 

Diff is long: https://github.com/getsentry/github-workflows/compare/v1...v2.8.1 (assuming this is what v1 to v2 does)
maybe the value was always wrong but didn't get used in v1?
new version has `      api-token: ${{ secrets.CI_DEPLOY_KEY }}`

vaind (changes requested):
No need to roll back, just fix the config, see my pr #172

vaind:
> @vaind seems like v2 breaks because of a token thing but v1 works.
> 
> failure:
> 
> > The workflow is not valid. .github/workflows/update-deps.yml (Line: 14, Col: 11): Secret api-token is required, but not provided while calling. .github/workflows/update-deps.yml (Line: 20, Col: 18): Invalid secret, api_token is not defined in the referenced workflow.
> 
> Diff is long: [getsentry/github-workflows@v1...v2.8.1](https://github.com/getsentry/github-workflows/compare/v1...v2.8.1) (assuming this is what v1 to v2 does) maybe the value was always wrong but didn't get used in v1? new version has ` api-token: ${{ secrets.CI_DEPLOY_KEY }}`
> 
> But repo has `CI_DEPLOY_KEY` (possibly is empty if I fucked it up somehow

No need to look at the diff between v1 and the latest v2. The release notes say what changed at as very first item https://github.com/getsentry/github-workflows/releases/tag/v2.0.0
> Rename api_token secret to api-token (https://github.com/getsentry/github-workflows/pull/21)

PR #172, opened 2023-12-20
ci: fix updater config

### LINKED ISSUE #170: ci: update-deps.yml is failing
### Description

After updating it to `@v2` the `api-token` seems to cause the workflow to fail / not even run

example: https://github.com/getsentry/sentry-kotlin-multiplatform/actions/runs/7270442573

vaind:
#161 updated to v2 but did not update the config

fixes #170

#skip-changelog

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -16,8 +16,7 @@ jobs:
       path: scripts/update-cocoa.sh
       name: Cocoa SDK
     secrets:
-      # If a custom token is used instead, a CI would be triggered on a created PR.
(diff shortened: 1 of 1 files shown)


buenaflor (approved):
ty!

buenaflor:
I'll continue the pr on the other branch since one of the e2e tests uses a secret from the repo which cannot be used in prs created from outside contribs

vaind:
> I'll continue the pr on the other branch since one of the e2e tests uses a secret from the repo which cannot be used in prs created from outside contribs

That needs fixing, I've created an issue.

Issue #178, opened 2024-01-24
isCrashedLastRun method


alexwhb:
### Problem Statement

I want to determine if my app crashed on the last run using the multiplatform library. I saw that this is possible in the iOS & Android libraries. It seems this is not currently possible. 

Really happy you guys made this library. It's honestly been really good.

PR #186, opened 2024-02-19
Add isCrashedLastRun

buenaflor:
## :scroll: Description

Adds `isCrashedLastRun`

**Note**: codecov ci fails because I haven't added any test, not sure what the best way to test here is since mocking on KMP is not really a thing yet

Closes https://github.com/getsentry/sentry-kotlin-multiplatform/issues/178

Manually

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Features
+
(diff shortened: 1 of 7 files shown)


romtsn (approved):
I think you can omit tests and rely on the native SDKs here, otherwise would've use a lot of internal stuff to test it probably

PR #188, opened 2024-02-20
ci: change macos runners

buenaflor:
use arm macos runners

#skip-changelog


romtsn (approved):
the builds are flying 🚀

PR #208, opened 2024-04-24
Make `setSentryUnhandledExceptionHook` public

### LINKED ISSUE #200: Make `setSentryUnhandledExceptionHook` public
### Problem Statement

We have a mixed KMP and native app, meaning that the limited configuration options available via this library, limit what we can enable on both the Android and iOS sides.

On the iOS side, AFAICT the only thing library does is called `SentrySDK` along with setting an unhandled exception handler. 

### Solution Brainstorm

Since I would rather call `SentrySDK.start {}` in Swift, can we make the [setSentryUnhandledExceptionHook](https://github.com/getsentry/sentry-kotlin-multiplatform/blob/main/sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/SentryUnhandledExceptions.kt#L52C14-L52C45) function public?

buenaflor:
## :scroll: Description

Makes the `setSentryUnhandledExceptionHook` api public so users can initialize the iOS SDK themselves and drop the unhandled exceptions created automatically that doesn't have useful information. See [here](https://github.com/rickclephas/NSExceptionKt?tab=readme-ov-file#why-this-library)

Closes #200 

X

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancements
+
+- Make `setSentryUnhandledExceptionHook` public ([#208](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/208))
+
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


krystofwoldrich (approved):
LGTM 🚀

PR #217, opened 2024-05-03
Fix: don't crash app when `applicationContext` is not available

### LINKED ISSUE #212: Problem when used together with Phoenix library
### Platform

Android

### Installed

None

### Version

0.5.0

### Steps to Reproduce

1. Use new init function that does not take `Context` and relies on `ContentProvider`
2. In the app at some point do `ProcessPhoenix.triggerRebirth(context)`

### Expected Result

App should not crash

### Actual Result

App crashes with
```
Caused by: kotlin.UninitializedPropertyAccessException: lateinit property applicationContext has not been initialized
	at io.sentry.kotlin.multiplatform.SentryInit_androidKt.getApplicationContext(SentryInit.android.kt:17)
	at io.sentry.kotlin.multiplatform.SentryInit_androidKt.initSentry(SentryInit.android.kt:14)
	at io.sentry.kotlin.multiplatform.SentryBridge.init(SentryBridge.jvm.kt:21)
	at io.sentry.kotlin.multiplatform.Sentry.init(SentryKMP.kt:43)
```

buenaflor:
## :scroll: Description

`lateinit` crashes the application if not set for example when using the Phoenix library

When calling `ProcessPhoenix.triggerRebirth(context)` the app is actually restarted twice and only one the second time the content providers are loaded.

Fixes #212 

Manual, `SentryContextProviderTest`

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryInit.android.kt
+++ b/sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryInit.android.kt
@@ -5,16 +5,24 @@ import android.content.ContentValues
 import android.content.Context
 import android.database.Cursor
 import android.net.Uri
+import android.util.Log
 import io.sentry.android.core.SentryAndroid
 import io.sentry.kotlin.multiplatform.extensions.toAndroidSentryOptionsCallback
 
 internal actual fun initSentry(configuration: OptionsConfiguration) {
     val options = SentryOptions()
     configuration.invoke(options)
-    SentryAndroid.init(applicationContext, options.toAndroidSentryOptionsCallback())
-}
 
-internal lateinit var applicationContext: Context
+    val context = applicationContext ?: run {
+        // TODO: add logging later
+        return
+    }
+
+    SentryAndroid.init(context) { sentryOptions ->
+        options.toAndroidSentryOptionsCallback().invoke(sentryOptions)
+    }}
+
+internal var applicationContext: Context? = null
     private set
 
 public actual typealias Context = Context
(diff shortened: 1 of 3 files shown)


sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryInit.android.kt
```diff
@@ -11,10 +11,18 @@ import io.sentry.kotlin.multiplatform.extensions.toAndroidSentryOptionsCallback
 internal actual fun initSentry(configuration: OptionsConfiguration) {
     val options = SentryOptions()
     configuration.invoke(options)
-    SentryAndroid.init(applicationContext, options.toAndroidSentryOptionsCallback())
+
+    val context = applicationContext ?: run {
+        // TODO: add logging later
+        return
```
romtsn, line 17: tbh, I am a bit hesitant with this approach, because if someone removes our `SentryContextProvider` for whatever reason (as mentioned [here](https://docs.sentry.io/platforms/android/configuration/manual-init/#manual-initialization) for example), the SDK init will silently fail and the users wont be aware of that. Perhaps there's a better way to handle ProcessPhoenix specifically?

romtsn: is the problem that ContentProviders are not loaded at all, or that `context` or `context.applicationContext` is null?

buenaflor: I've set up some debug logs and the application init is triggered twice and during the the content provider is not loaded before so applicationContext crashes due to `lateinit` but on the second load it works fine because the content provider has already set the context:

```
Rebirth triggered now (click rebirth button)
Try init sentry android: app context null (app tries to restart process, content provider is not triggered before)
Content provider loaded (now content provider is triggered)
Try init sentry android: app context sample.kmp.app.android.SentryApplication@d0c4ab0 (now we have access to app context)
```

romtsn: ok, I see, I guess there's no many ways around it anyway. After we implement logging that should help also 👍

romtsn (approved):

Issue #213, opened 2024-05-02
Expose platform specific Configuration via expect actual


Legion2:
### Problem Statement

I want to configure underlying configurations in android and ios SDK, which are not jetzt available in the common configuration.

### Solution Brainstorm

Create an expected type for the configuration in common code and add actual implementations of it in the platforms. The users can then also use expect/actual to configure the actual platform configuration as needed, without there being a common mapping between them.

Legion2:
We would like to try out the React Native session replay alpha (https://github.com/getsentry/sentry-react-native/issues/2703), but can not apply the required configuration in sentry android and ios SDK.

buenaflor:
> Create an expected type for the configuration in common code and add actual implementations of it in the platforms. The users can then also use expect/actual to configure the actual platform configuration as needed, without there being a common mapping between them.

can you give a simple example?

Legion2:
```kotlin
// common
expect class PlatformConfigurationOptions

// android
actual typealias PlatformConfigurationOptions = SentryAndroidOptions // class from android sdk

// ios
actual typealias PlatformConfigurationOptions = SentryiOSOptions // class from ios sdk
```

buenaflor:
Okay understand

This should work, theoretically, but testing it, there's some issue with typealiasing the JVM sentryoptions and Android sentry options where you cannot access the properties. have you been able to successfully implement this locally?

buenaflor:
Odd thing is that it works directly in the SDK but when consumed by another application (e.g our sample app) it fails

buenaflor:
Okay got it why it didn't work, it requires us to expose the internal native API's with `api('java_sdk_dependency)`. 

It is possible and simplifies the customization of the native SDK tbh. We generally haven't used `actual typealias` yet because it requires exposing the internal Sentry Native SDK dependencies.

Tbh might be worth it but I'll talk to my team and see if we have any concerns about that

Legion2:
If you don't want to expose the dependency on the android SDK, you can specify it in the documentation, that users that want to use the platform config should add the dependency on the android SDK themselves to their projects.

buenaflor:
I'm generally thinking about exposing them since it would reduce the internal jvm/android implementation side immensely if we can internally use typealias as much as possible as well

But good point, regardless we can document that in the docs

buenaflor:
[...]

PR #221, opened 2024-05-10
feat: override native options

buenaflor:
## :scroll: Description
Allows user to init with `initWithPlatformOptions { }` in the respective platform sourceSet which directly calls the native options

Requirements for being able to use this:
- using the Gradle cocoapods plugin to create bindings for apple targets

Also open to changes to the API name

(the implementation is a pretty simple addition but changing multiple platforms in KMP just blows up the files changed stats)

Implements #213 

Unit tests

## :pencil: Checklist

## :crystal_ball: Next steps


sentry-kotlin-multiplatform/src/androidMain/kotlin/io/sentry/kotlin/multiplatform/SentryPlatformOptions.android.kt
```diff
@@ -0,0 +1,17 @@
+package io.sentry.kotlin.multiplatform
+
+import io.sentry.kotlin.multiplatform.extensions.toAndroidSentryOptionsCallback
+
+public actual typealias SentryPlatformOptions = io.sentry.android.core.SentryAndroidOptions
+
+internal actual fun SentryPlatformOptions.prepareForInit() {
+    sdkVersion?.name = BuildKonfig.SENTRY_KMP_ANDROID_SDK_NAME
+    sdkVersion?.version = BuildKonfig.VERSION_NAME
+}
```
romtsn, line 10: should we also move [these lines](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/221/files#diff-14b13973931d1582b7429057cdb84ba6068f69479da997c3c8f39a7cff12459dR24-R26) over here?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add feature to initialize with native options ([#221](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/221))
```
romtsn, line 7:
```suggestion
- Allow initializing SentryKMP SDK with native options ([#221](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/221))
```
I feel like "feature" is excessive here, since it's under the "Features" section already. Also, can we add a code snippet here to showcase how users can provide platform options?

buenaflor: yup added code snippet as well

romtsn (approved):
LGTM!

anything against making the native dependencies `api`? I think that's what we do on RN for example

buenaflor:
@romtsn actually I'd love to add it, then we can freely typealias stuff directly to the android/java sdk and the user doesn't need to add `compileOnly(...)` in their apps

romtsn:
> @romtsn actually I'd love to add it, then we can freely typealias stuff directly to the android/java sdk and the user doesn't need to add `compileOnly(...)` in their apps
> 
> the only thing speaking against it is that you'd have access to the android/java sdk directly in your android sourceSet but not sure if that's really so bad
> 
> The user just needs to be careful not to confuse `io.sentry.Sentry` with `io.sentry.kotlin.multiplatform.Sentry` for example

Let's expose it, I think we can always get back and re-think if there are problems. I also expect problems for those who use multiple sentry SDKs (e.g. RN and KMP which both expose the Android sdk), but then it's kind of not our problem and they'd have to resolve conflicts themselves. Just have to make sure we're not lagging behind with SDK updates for major bumps/breaking changes

Issue #191, opened 2024-02-29
Better support for tests not working on cocoa targets


buenaflor:
### Description

Cocoa tests will not work in the following scenarios: 
- using SPM
- packages that consume a shared package where cocoapods is set up. Example:
    - Package A is a package that is consumed by multiple other packages (maybe a package that deals with network requests).
    - Package B is the main shared package that is set up with Cocoapods and package B consumes package A as a dependency.
        - Assumption by users: since Package A is already set up with cocoapods, it will transitively work for tests in package B → it does not
        - Reason: since package B is isolated in its own package it has no reference as to where to get the binary for the cocoa SDK so naturally it will fail
        
## Solutions
- Temporary workaround can be found [here](https://docs.sentry.io/platforms/kotlin-multiplatform/swift-package-manager/#running-tests).
- Ideally we add the temporary workaround impl to a gradle plugin
- Provide gradle tasks code snippets that users can just copy and paste

let's decide how we wanna deal with this going forward

PR #230, opened 2024-06-05
feat: add plugin support

### LINKED ISSUE #108: [Gradle Plugin] Add support for creating a dynamic library when using SPM and Cocoapods (not kotlin cocoapods)
### Description

Currently if someone wants to use SPM or (native, not kotlin) Cocoapods with our SDK it is only possible when making the shared libary static (dynamic is the default), e.g: 

```kotlin
kotlin {
  listOf(
    iosX64(),
    iosArm64(),
    iosSimulatorArm64()
  ).forEach {
    it.binaries.framework {
        baseName = "shared"
        isStatic = true // Using SPM is only available when creating a static framework
    }
  }
```

### Why the current solution is not ideal

- The sure way to get dynamic framework to work is by using the kotlin cocoapods plugin to pull in the sentry-cocoa dependency, however that is inefficient as it runs through c-interop which also creates extra compile time and overhead in the final binary (also probably a lot of dead code)

### Advantages of having it work as a dynamic library

- SwiftUI previews don't work with static libraries. There might be some workaround but that adds extra layers of complexity for the user.

### Why it doesn't work

On iOS only the definitions for Sentry are added when compiling the Kotlin code and building the framework. The binary (the actual Sentry Cocoa library) isn't added until later when you build the iOS app. When building a dynamic framework, the Kotlin compile expects to be able to resolve everything and it ends up with this error:
`ld: symbol(s) not found for architecture x86_64` which basically tells us it cannot find the binary. 

### Solution

~[Crashkios](https://github.com/touchlab/CrashKiOS/blob/main/crashlytics-ios-link/src/main/kotlin/co/touchlab/crashkios/CrashlyticsLinkPlugin.kt) has solved this by basically telling the linker to treat the listed symbols as undefined. That means the linker will not try to find these symbols during the linking process and assumes these symbols will be provided by other means.~

The solution might not be viable since it can lead to problems if the app has app thinning enabled. See https://youtrack.jetbrains.com/issue/KT-58461

An alternative solution would be to properly link it to the framework

Essentially this means creating a KMP gradle plugin.

```[tasklist]
### Blocked by
- [ ] https://github.com/getsentry/sentry-kotlin-multiplatform/issues/191
```

buenaflor:
## :scroll: Description

Adds plugin for easier installation process

Fixes #108 and https://github.com/getsentry/sentry-kotlin-multiplatform/issues/191

Unit tests, samples

## :pencil: Checklist

## :crystal_ball: Next steps


buildSrc/src/main/java/Config.kt
```diff
@@ -1,7 +1,7 @@
 object Config {
     val agpVersion = "7.4.2"
-    val kotlinVersion = "1.9.21"
-    val composeVersion = "1.5.11"
+    val kotlinVersion = "1.9.23"
+    val composeVersion = "1.6.1"
```
buenaflor, line 4: this is only temporary, with the current xcode version compiling with Kotlin <= 1.9.23 doesn't work

buenaflor:
@romtsn I haven't added any tests, just wanted to see what you think about the current state of this draft, the important files I'd say are `LinkerExtension.kt`, `SentryExtension.kt` and `SentryPlugin.kt`

the rest hasn't really been polished or anything yet

buenaflor:
also tbd: auto installing the kmp dependency in commonMain

plugin-build/build.gradle.kts
romtsn, line 1: I know you took it from a template, but i'd prefer renaming `plugin-build` folder to something more descriptive like `sentry-kotlin-multiplatform-gradle-plugin`

sentry-kotlin-multiplatform/build.gradle.kts
```diff
@@ -145,7 +145,10 @@ kotlin {
             homepage = "https://github.com/getsentry/sentry-kotlin-multiplatform"
             version = "0.0.1"
 
-            pod(Config.Libs.sentryCocoa, Config.Libs.sentryCocoaVersion)
+            pod(Config.Libs.sentryCocoa) {
+                version = Config.Libs.sentryCocoaVersion
+                extraOpts += listOf("-compiler-option", "-fmodules")
```
romtsn, line 150: `l`: also, is this needed now?

sentry-samples/kmp-app-cocoapods/iosApp/iosApp.xcworkspace/xcshareddata/swiftpm/Package.resolved
romtsn, line 1: also feels like a leftover, maybe should be ignored?

plugin-build/gradle.properties
```diff
@@ -0,0 +1,8 @@
+ID=io.sentry.kotlin.multiplatform.gradle.plugin
```
romtsn, line 1: nit: for android we just use `io.sentry.android.gradle`, so maybe we could drop the last `.plugin` part here as well

plugin-build/gradle.properties
```diff
@@ -0,0 +1,8 @@
...
+VERSION=1.0.0
```
romtsn, line 2: i see this is still from the template, I guess you're gonna change, just don't forget to align the version with the SDK version itself as we've agreed

plugin-build/plugin/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+
+plugins {
+    kotlin("jvm")
+    `java-gradle-plugin`
+    alias(libs.plugins.vanniktech.publish)
+    id("distribution")
+}
+
+dependencies {
+    implementation(kotlin("stdlib"))
```
romtsn, line 12: `m`: I think these all should be `compileOnly` otherwise we'd be bundling kgp and gradle-api with our plugin and potentially overwriting what users use in their build

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/CocoapodsAutoInstallExtension.kt
```diff
@@ -0,0 +1,32 @@
...
+                .convention(true)
+
+        /**
+         * Overrides default Sentry Cocoa version.
+         *
+         * Defaults to the `~> 8.25`.
+         */
+        val sentryCocoaVersion: Property<String> =
+            objects.property(String::class.java)
+                .convention("~> 8.25") // todo: grab the latest version properly
```
romtsn, line 31: Could do it e.g. via `buildConfig` fields, similar to what we do for the sentry SDK in SAGP -  store it in `gradle.properties` and then retrieve like this https://github.com/getsentry/sentry-android-gradle-plugin/blob/f9b82ddce3af1a21eec54da8b1c02e36c063939a/plugin-build/build.gradle.kts#L348

settings.gradle.kts
```diff
@@ -20,6 +20,7 @@ dependencyResolutionManagement {
 rootProject.name = "sentry-kotlin-multiplatform-sdk"
 
 include(":sentry-kotlin-multiplatform")
+includeBuild("plugin-build")
```
romtsn, line 23: nit: I think you'd need to define `dependencySubstitution` block so you don't have to publish to local maven every time, and just develop against the included build, like here: https://github.com/getsentry/sentry-android-gradle-plugin/blob/f9b82ddce3af1a21eec54da8b1c02e36c063939a/settings.gradle.kts#L37-L41

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
```diff
@@ -0,0 +1,231 @@
...
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.jetbrains.kotlin.gradle.plugin.cocoapods.CocoapodsExtension
+import org.jetbrains.kotlin.gradle.plugin.cocoapods.KotlinCocoapodsPlugin
+import org.jetbrains.kotlin.gradle.plugin.mpp.Framework
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.TestExecutable
+import java.io.ByteArrayOutputStream
+import java.io.File
+
+internal const val SENTRY_EXTENSION_NAME = "sentry"
```
romtsn, line 16: `m`: I'm wondering if we should consider the case where both this plugin and SAGP are used. In that case there will probably a conflict of two extensions with the same name, so perhaps we should consider something different here, like `sentryKmp` or `sentryMultiplatform`

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
```diff
@@ -0,0 +1,231 @@
...
+    val unsupportedTargets = listOf("wasm", "js", "mingw", "linux")
+    kmpExtension.targets.forEach { target ->
+        if (unsupportedTargets.any { unsupported ->
+                target.name.contains(unsupported)
+            }
+        ) {
+            throw GradleException(
+                "Unsupported target: ${target.name}. " +
+                    "Cannot auto install in commonMain. " +
+                    "Please create an intermediate sourceSet with targets that the SDK supports and add the dependency manually.",
```
romtsn, line 82:
```suggestion
                    "Please create an intermediate sourceSet with targets that the Sentry SDK supports (apple, jvm, android) and add the dependency manually.",
```

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
```diff
@@ -0,0 +1,231 @@
...
+}
+
+internal fun Project.installSentryForKmp(commonMainAutoInstallExtension: SourceSetAutoInstallExtension) {
+    val kmpExtension = extensions.findByName(KOTLIN_EXTENSION_NAME)
+    if (kmpExtension !is KotlinMultiplatformExtension) {
+        // todo: log, not multiplatform found
+        return
+    }
+
+    val unsupportedTargets = listOf("wasm", "js", "mingw", "linux")
```
romtsn, line 73: there's also androidNative which I guess we also don't support through the KMP sdk, right? https://kotlinlang.org/docs/native-target-support.html#tier-3

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
```diff
@@ -0,0 +1,231 @@
...
+    val buildDirOutput = ByteArrayOutputStream()
+    exec {
+        it.commandLine =
+            listOf(
+                "xcodebuild", "-project", xcodeprojPath, "-showBuildSettings",
+            )
+        it.standardOutput = buildDirOutput
+    }
+    val buildSettings = buildDirOutput.toString("UTF-8")
+    val buildDirRegex = Regex("BUILD_DIR = (.+)")
```
romtsn, line 174: `m`: maybe worth to extract this into a companion object, so it's only computed once

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
```diff
@@ -0,0 +1,231 @@
...
+            "$derivedDataPath/SourcePackages/artifacts/sentry-cocoa/Sentry/Sentry.xcframework/$frameworkArchitecture"
+
+        target.binaries.all binaries@{ binary ->
+            val path =
+                when {
+                    File(dynamicFrameworkPath).exists() -> dynamicFrameworkPath
+                    File(staticFrameworkPath).exists() -> {
+                        // todo: log, dynamic framework not found, try using static framework
+                        staticFrameworkPath
+                    }
```
romtsn, line 146: `h`: iirc, if we link the static framework to a dynamic kmp library, it will crash at runtime right? can/should we also check if the binary `isStatic` in this case? Basically have supported combinations of framework-klib

romtsn (commented):
I think there's a couple of comments to be addressed, but the rest is minor stuff. Looks good otherwise, I think this is dope!

sentry-samples/kmp-app-cocoapods/shared/build.gradle.kts
```diff
@@ -31,7 +32,10 @@ kotlin {
         ios.deploymentTarget = "14.1"
         podfile = project.file("../iosApp/Podfile")
 
-        pod("Sentry", Config.Libs.sentryCocoaVersion)
+        pod("Sentry") {
+            version = Config.Libs.sentryCocoaVersion
+            extraOpts += listOf("-compiler-option", "-fmodules")
```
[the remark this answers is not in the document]
buenaflor: should be up to date now, just needed to merge main

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
buenaflor: yeah definitely

.github/workflows/kotlin-multiplatform-gradle-plugin.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          java-version: 17
+          distribution: temurin
+
+      - name: Cached Gradle
+        uses: gradle/gradle-build-action@842c587ad8aa4c68eeba24c396e15af4c2e9f30a
+
+      - name: DistZip
+        run: |
+          cd sentry-kotlin-multiplatform-gradle-plugin && ./gradlew distZip
```
romtsn, line 31: not sure if `distZip` runs tests, but I guess we also should run tests? Or you wanted to have a matrix later on?

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add plugin ([#230](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/230))
```
romtsn, line 7:
```suggestion
- New Sentry KMP Gradle plugin ([#230](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/230))
```

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/CocoapodsAutoInstallExtension.kt
```diff
@@ -0,0 +1,29 @@
...
+     */
+    val enabled: Property<Boolean> = objects.property(Boolean::class.java).convention(true)
+
+    /**
+     * Overrides default Sentry Cocoa version.
+     *
+     * Defaults to the version used in the latest KMP SDK.
+     */
+    val sentryCocoaVersion: Property<String> =
+        objects.property(String::class.java).convention(BuildConfig.SentryCocoaVersion)
```
romtsn, line 28: should we prepend this with that weird arrow `~>` here actually?

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/SourceSetAutoInstallExtension.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Overrides the default Sentry Kotlin Multiplatform SDK dependency version.
+     *
+     * Defaults to the latest version of the Sentry Kotlin Multiplatform SDK.
+     */
+    val sentryKmpVersion: Property<String> =
+        objects
+            .property(String::class.java)
+            .convention("latest.release") // Replace with the actual default version
```
romtsn, line 27: I think we could also already use the `BuildConfig` value here which we could feed from `gradle.properties/versionName`

settings.gradle.kts
buenaflor: it already works even without `dependencySubstitution`. 

I don't need to do a local publish during dev

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
buenaflor: moved it as a top level declaration

plugin-build/plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/plugin/SentryPlugin.kt
```diff
@@ -0,0 +1,231 @@
...
+                binary.linkerOpts("-F$path")
+            }
+        }
+    }
+}
+
+private fun Project.findDerivedDataPath(customXcodeprojPath: String? = null): String {
+    val xcodeprojPath = customXcodeprojPath ?: findXcodeprojFile(rootDir)?.absolutePath
+    val buildDirOutput = ByteArrayOutputStream()
+    exec {
```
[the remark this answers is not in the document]
buenaflor: Added fetching the derived data path as a value source

settings.gradle.kts
romtsn: oh, ok, I didn't know that works! probably when using the new `plugins { }` block, it just works

romtsn (approved):
I think this is good to go, great work!

buenaflor:
@romtsn 

just gonna add a couple more tests.

should we do an alpha or beta first?

romtsn:
> @romtsn
> 
> just gonna add a couple more tests.
> 
> should we do an alpha or beta first?

yeah, I think a beta is reasonable, maybe ping people on the issues that were reported to try this out as well. Or add to the docs, like a banner or whatever

PR #250, opened 2024-08-01
Update Kotlin version to `1.9.23`

buenaflor:
There is a known issue with compiling on xcode `15.3` that is fixed only on Kotlin `1.9.23` and higher.

We are on `1.9.21` and I have to change it to `1.9.23` every time when developing on KMP.

See: https://youtrack.jetbrains.com/issue/KT-65542/Cinterop-tasks-fails-if-Xcode-15.3-is-used

I'm not sure about the backwards compatibility implications but I think it's fine?

@romtsn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


buenaflor:
I think this change is generally necessary since the default xcode version on github is now `15.4`

romtsn:
I think it's fine, just a patch version

romtsn (approved):

PR #252, opened 2024-08-08
fix: plugin publication

buenaflor:
previously plugin and library workflows did the uploading of the distributions separately, I don't think this would've worked.

Now the distributions are packaged in one workflow.

Also added a couple things like the pom metadata

#skip-changelog


sentry-kotlin-multiplatform-gradle-plugin/build.gradle.kts
```diff
@@ -52,32 +52,39 @@ val publish = extensions.getByType(MavenPublishPluginExtension::class.java)
...
     }
 }
 
+tasks.named("sentryPluginMarkerDistTar") {
+    mustRunAfter("generatePomFileForSentryPluginPluginMarkerMavenPublication")
+}
+
+tasks.named("sentryPluginMarkerDistZip") {
+    mustRunAfter(":generatePomFileForSentryPluginPluginMarkerMavenPublication")
+}
```
buenaflor, line 86: somehow I need this or it gives me a "without declaring an explicit or implicit dependency." error

.github/workflows/upload-artifacts.yml
```diff
@@ -0,0 +1,42 @@
...
+          cd sentry-kotlin-multiplatform-gradle-plugin
+          ./gradlew distZip sentryPluginMarkerDistZip
+
+      - name: Archive packages
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ github.sha }}
+          if-no-files-found: error
+          path: |
+            ./*/build/distributions/*.zip
```
buenaflor, line 42: merge all the distributions into one single workflow

buenaflor:
@romtsn I think after this PR the release will work fine

romtsn (approved):

PR #265, opened 2024-09-06
refactor: improve interop with objc headers

### LINKED ISSUE #74: Use trimmed headers instead of cocoapods for Apple targets
I decided to keep using cocoapods because of several reasons:

- When I started 'rebuilding' a couple classes in custom objc headers such as SentryEvent I noticed how many classes we actually need and at some point we would maybe shave off 4-5 files (rough guess, cant remember exactly now) if we did it all by hand, which is not a lot.
- The total size of the klib file which contains the bindings that were created by the cocoapods plugin is around 70kb
- Future maintenance just sounds awful when we have to manually add/remove stuff when things are deprecated or added and it's easy to make a mistake thus crashing the users app
- Modularized sentry-cocoa may come soon-ish

What we should do: combine all the cinterop libraries that we had into one `SentryInternal` and shave off a lot of objc headers that aren't needed

buenaflor:
## :scroll: Description

Combine the cinterop libraries that we had into one, remove unnecessary header files thus reducing klib file size

Closes #74 

## :pencil: Checklist

- [x] ~I added tests to verify the changes.~

## :crystal_ball: Next steps


sentry-kotlin-multiplatform/build.gradle.kts
```diff
@@ -177,14 +177,8 @@ kotlin {
...
             it.compilations.getByName("main") {
-                cinterops.create("Sentry.NSException") {
-                    includeDirs("$projectDir/src/nativeInterop/cinterop/SentryNSException")
-                }
-                cinterops.create("Sentry.Scope") {
-                    includeDirs("$projectDir/src/nativeInterop/cinterop/SentryScope")
-                }
-                cinterops.create("Sentry.PrivateSentrySDKOnly") {
-                    includeDirs("$projectDir/src/nativeInterop/cinterop/SentryPrivateSentrySDKOnly")
+                cinterops.create("Sentry.Internal") {
```
romtsn, line 180: nit: I think we can remove the workaround below (line 188) now, because we don't include those headers anymore, wdyt?

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/SentryUnhandledExceptions.kt
```diff
@@ -67,56 +71,60 @@ internal const val KOTLIN_CRASH_TAG = "nsexceptionkt.kotlin_crashed"
...
-        stacktrace = null
+private fun Throwable.asSentryEvent(): CocoapodsSentryEvent =
+    CocoapodsSentryEvent(kSentryLevelFatal).apply {
+        @Suppress("UNCHECKED_CAST")
+        val threads =
+            threadInspector?.getCurrentThreadsWithStackTrace() as List<CocoapodsSentryThread>?
+        this.threads = threads
+        val currentThread = threads?.firstOrNull { it.current?.boolValue ?: false }?.apply {
+            setCrashed(NSNumber(true))
+            // Crashed threats shouldn't have a stacktrace, the thread_id should be set on the exception instead
```
romtsn, line 98:
```suggestion
            // Crashed threads shouldn't have a stacktrace, the thread_id should be set on the exception instead
```

romtsn (approved):
LGTM! 

I'm wondering how much more we could shave off by implementing this in the Cocoa SDK https://github.com/getsentry/sentry-cocoa/issues/1999. Probably not that much work tbh, but we could do it as a next step improvement later

sentry-kotlin-multiplatform/build.gradle.kts
buenaflor: we still do, the cocoapods plugin is still in use so it also pulls in `SentryMechanismMeta`

I decided to keep using cocoapods because of several reasons:
- When I started 'rebuilding' a couple classes in custom objc headers such as SentryEvent I noticed how many classes we actually need and at some point we would maybe shave off 4-5 files if we did it all by hand, which is not a lot.
- The total size of the klib file which contains the bindings that were created by the cocoapods plugin is around 70kb 
- Future maintenance just sounds awful when we have to manually add/remove stuff when things are deprecated or added and it's easy to make a mistake thus crashing the users app

What I did instead: combine all the cinterop libraries that we had into one `SentryInternal` and I shaved off a lot of objc headers that wasn't needed

romtsn: ah ok, gotchu 👍

PR #276, opened 2024-09-30
ci: remove auto bot format

buenaflor:
The bot will run auto formatting but the CI will freeze because bots don't have permission to run CI so someone has to re-trigger the CI with another change which is annoying.

The analyze workflow will fail anyway if formatting is off, it's more convenient at this point to just format locally and then push manually

#skip-changelog


romtsn (approved):
I agree with the change, but ideally we'd fix the bot somehow that it does not require a retrigger 😅 like using bruno-garcia's token lol

PR #283, opened 2024-10-09
fix(publish): regex doesn't match double digit releases

buenaflor:
regex in `craft.yml` doesn't match releases with double digits e.g `0.10.0` which would replace the aar name

#skip-changelog


romtsn (approved):
nice touch with the validate task 👍 also I feel you :D https://github.com/getsentry/sentry-java/pull/2433

buenaflor:
> also I feel you :D https://github.com/getsentry/sentry-java/pull/2433

ahh lol 🤣

PR #296, opened 2024-11-19
chore: Update PR template

philipphofmann:
Remove I reviewed the code and add updating the wizard and the docs because we often forget to update these.

#skip-changelog


buenaflor:
We don't have wizard support yet for KMP

philipphofmann:
Oh, that's sad. Closing it then.

PR #320, opened 2025-01-30
fix(gradle-plugin): architecture folder name missmatch when using SPM and framework path searching

buenaflor:
## :scroll: Description

Fixes several things
- not finding the correct architecture in newer cocoa versions xcframeworks
- `frameworkPath` doesn't check for architecture folders
- add multiple strategies for finding the cocoa framework path when using SPM, if one fails to find it, the next one is triggered:
  - 1. custom framework path set by user `frameworkPath`
  - 2. derived data using the xcodeproj with `xcodebuild -showBuildSettings`
  - 3. searching derived data manually using commands

Right now `8.38.0` and above won't work with the plugin with SPM

Added test that downloads the frameworks and tests against it.

## :pencil: Checklist

## :crystal_ball: Next steps


sentry-kotlin-multiplatform-gradle-plugin/src/test/java/io/sentry/kotlin/multiplatform/gradle/Old.kt
```diff
@@ -0,0 +1,140 @@
+package io.sentry.kotlin.multiplatform.gradle
```
romtsn: Do you wan to drop this class or leave a TODO here?

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/FrameworkLinker.kt
```diff
@@ -0,0 +1,59 @@
...
+            when (binary) {
+                is TestExecutable -> linkTestBinary(binary, chooseTestPath(dynamicPath, staticPath))
+                is Framework -> linkFrameworkBinary(binary, dynamicPath, staticPath)
+                else -> logger.info("Ignoring binary type: ${binary::class.java.simpleName}")
+            }
+        }
+    }
+
+    private fun chooseTestPath(dynamic: String?, static: String?) = when {
+        dynamic != null && static != null -> {
```
romtsn, line 30: do you really need this condition? could as well just omit it, and it'd always pick `dynamic` when it's available or fall back to static otherwise

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/FrameworkPathResolver.kt
```diff
@@ -0,0 +1,256 @@
...
+        ): FrameworkPaths {
+            val dynamicPath = dynamicBasePath?.let { basePath ->
+                architectures.map { arch -> "$basePath/$arch" }.firstOrNull { pathExists(it) }
+            }
+
+            val staticPath = staticBasePath?.let { basePath ->
+                architectures.map { arch -> "$basePath/$arch" }.firstOrNull { pathExists(it) }
+            }
+
+            return when {
```
romtsn, line 32: also can be simplified to

```
if (dynamicPath == null && staticPath == null) NONE else FrameworkPaths(dynamic = dynamicPath, static = staticPath)
```

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/FrameworkPathResolver.kt
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * Searches for a xcodeproj starting from the root directory. This function will only work for
+     * monorepos and if it is not, the user needs to provide the [LinkerExtension.xcodeprojPath].
+     */
+    private fun findXcodeprojFile(startingDir: File): File? {
+        val ignoredDirectories = listOf("build", "DerivedData")
+
+        fun searchDirectory(directory: File): File? {
+            val files = directory.listFiles() ?: return null
```
romtsn, line 135: I guess we could also use `walkFileTree` or something like that from kotlin, but this is also fine like it's done now

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/FrameworkPathResolver.kt
```diff
@@ -0,0 +1,256 @@
...
+                if (result != FrameworkPaths.NONE) {
+                    val path = result.dynamic ?: result.static
+                    project.logger.lifecycle("Found Sentry Cocoa framework path using ${strategy::class.simpleName} at $path")
+                    return result
+                } else {
+                    project.logger.debug("Strategy ${strategy::class.simpleName} did not find valid paths")
+                }
+            } catch (e: FrameworkLinkingException) {
+                project.logger.warn(
+                    "Strategy ${strategy::class.simpleName} failed due to error: ${e.message}"
```
romtsn, line 213: there's a `warn` overload that accepts `Throwable` as a parameter - should we use it here to pass the exception to have it print the stacktrace nicely?

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/ManualFrameworkPathSearchValueSource.kt
```diff
@@ -0,0 +1,61 @@
...
+                    "-exec stat -f \"%m %N\" {} \\; | " +
+                    "sort -nr | " +
+                    "cut -d' ' -f2-"
+            )
+            it.standardOutput = output
+            it.isIgnoreExitValue = true
+        }
+
+        val stringOutput = output.toString("UTF-8")
+        return if (stringOutput.lineSequence().firstOrNull().isNullOrEmpty()) {
```
romtsn, line 54: I think we should also account for `exitValue != 0`, if e.g. the find command fails/doesn't exist/etc. You can check how we do it here (and also could pipe the error out stream to later print it out for easier debugging): https://github.com/getsentry/sentry-android-gradle-plugin/blob/14a69bcdc41d00f01fa8fdd7fa7648715da89109/plugin-build/src/main/kotlin/io/sentry/android/gradle/telemetry/SentryTelemetryService.kt#L484-L527

romtsn: it's also because you're just getting the first line in the output here, but it could be `Failed to execute 'find' ...` or something like that, and we don't do any validation over it as far as I can see

romtsn (approved):
This PR is very well done, hats off! 

I really like how you split the search up into different strategies. The use of parameterized tests is :chefkiss: and also the fact that you're downloading a real xcframework from github is making a perfect integration test!

Just left a few minor comments, but nothing big apart from handling exitValue maybe.

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/FrameworkPathResolver.kt
buenaflor: ah yeah thats a better api to use I think, I'll have a look at it

buenaflor: changed to using `walkFileTree` instead

PR #367, opened 2025-04-16
Move replay options out of experimental

### LINKED ISSUE #348: Session Replay GA
### Problem Statement

SR is currently available as experimental option, so [users on 0.10.0+ can enable it](https://github.com/getsentry/sentry-kotlin-multiplatform/blob/main/CHANGELOG.md#0100), but it’s not recommended for production use. 

For iOS for example, we need to make sure that masking works (or document where it doesn't, e.g. with skia rendering).
KMP also doesn’t use the latest Android/iOS SDKs yet, which have a number of bugfixes for SR in them.

### Solution Brainstorm

- update to stable iOS/Android SDK releases with GA SR support
- document how to use SR with KMP and possible limitations

buenaflor:
Closes #348 

## :pencil: Checklist

## :crystal_ball: Next steps


maxkomarychev:
Hi 👋 .

What's the ETA to get this released as a new version? I'm currently unable to use latest ios sdk (8.49.0) with KMM package 0.11.0 due to the crash and I'm not able to downgrade ios sdk below 8.46.0 because of the compatibility with xcode 16.3.

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[Sentry.SentryExperimentalOptions setSessionReplay:]: unrecognized selector sent to instance 0x60000000d650'
```

buenaflor:
@maxkomarychev  hey, at the latest end of this week. Sorry for the inconvenience. 

Meanwhile if you cannot work with anything at all I'd suggest using something like https://www.xcodes.app/ to manage xcode versions so you can downgrade temporarily

PR #376, opened 2025-04-23
fix: beforeSend overriding default release even if it was not set

### LINKED ISSUE #337: On iOS default release version is cleared when no release is set on the Kotlin Multiplatform side.
### Platform

Apple

### Installed

Swift Package Manager

### Version

8.42.0

### Steps to Reproduce

1. Do not set `release` property in `SentryOptions` in the shared code on Kotlin Multiplatform side
2. Set properly `CFBundleIdentifier`, `CFBundleShortVersionString`, and `CFBundleVersion` in `info.plist` in the xcode project used with kotlin multi-platform
3. Send any kind of event to Sentry (including a crash)
4. When checking event / issue it has no associated release

### Expected Result

According to the kotlin multiplatform SDK [documentation](https://docs.sentry.io/platforms/kotlin/guides/kotlin-multiplatform/configuration/options/):

> [release](https://docs.sentry.io/platforms/kotlin/guides/kotlin-multiplatform/configuration/options/#release)
> Sets the release. Some SDKs will try to automatically configure a release out of the box

Then according to the iOS SDK [documentation](https://docs.sentry.io/platforms/apple/configuration/releases/#setting-a-release):

> [Setting a Release](https://docs.sentry.io/platforms/apple/configuration/releases/#setting-a-release)
> If no release name is set, the SDK creates a default combined from CFBundleIdentifier, CFBundleShortVersionString, and CFBundleVersion, for example my.project.name@2.3.12+1234.

Which implies that when no `release` is set up on the Kotlin Multiplatform side that value will default to whatever platform's SDK behaviour.
In case of iOS that should be a value based on `CFBundleIdentifier`, `CFBundleShortVersionString`, and `CFBundleVersion` values.

### Actual Result

Right now the `release` value transmitted to the sentry backend is `null` by default.

This is most likely caused by overriding the platform's default value here:

https://github.com/getsentry/sentry-kotlin-multiplatform/blob/f22d5c9487aeebab02728c097845d2b4b656b4ad/sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryEventExtensions.apple.kt#L13

without checking if there is any value in `kmpEvent` as it is done for some other fields:

https://github.com/getsentry/sentry-kotlin-multiplatform/blob/f22d5c9487aeebab02728c097845d2b4b656b4ad/sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/extensions/SentryEventExtensions.apple.kt#L8

buenaflor:
## :scroll: Description

We need to keep the KMP event model in sync with the native one. Users mainly edit events in KMP hooks like beforeSend, so we must track whether each field was actually modified:
If a field was changed, propagate the new value to the native event.• If it was untouched, leave the native default in place.
Previously we overwrote native defaults with null values by default e.g., Cocoa’s auto‑detected release was not working since we set a null release

Fixes #337 

TBD unit tests
## :pencil: Checklist

## :crystal_ball: Next steps


sentry-kotlin-multiplatform/src/commonMain/kotlin/io/sentry/kotlin/multiplatform/util/ApplyIfChanged.kt
```diff
@@ -0,0 +1,5 @@
+package io.sentry.kotlin.multiplatform.util
+
+internal inline fun <T> applyIfChanged(old: T, new: T, applyNew: (T) -> Unit) {
+    if (old != new) applyNew(new)
```
romtsn, line 4: hm, how does this work for non-primitive types? E.g. Breadcrumbs list or User? They have to be `data classes` or implement `equals` right?

buenaflor: yes good point, I'll double check which fields are affected by this

buenaflor: in any case we need some way to detect whether a value changed or not and only then apply it

buenaflor:
I'll update this PR, let's fix it only for the release field now. I'm not even sure if the other fields need to be checked like that as well

sentry-kotlin-multiplatform/src/appleTest/kotlin/io/sentry/kotlin/multiplatform/ApplyKmpEventTest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+        nativeEvent.applyKmpEvent(kmpEvent)
+
+        assertFalse(nativeEvent.wasReleaseSet)
+    }
+}
+
+// init with a SentryLevel as a workaround for:
+// Unable to call non-designated initializer as super constructor
+private class FakeSentryEvent : CocoaSentryEvent(Internal.Sentry.kSentryLevelFatal) {
```
buenaflor, line 21: can only test cocoa like this since the `SentryEvent` in java is final and we don't use a mocking framework yet

buenaflor:
@romtsn is this good to go?

romtsn (approved):

PR #389, opened 2025-05-16
Deps: Update Kotlin to `2.1.21`

buenaflor:
Kotlin `2.1.21` adds support for Xcode 16.3 and higher so this is important

Also update the compose version for the sample apps


buenaflor:
Setting it to draft, lots of problems with the K2 compiler right now

buenaflor:
@romtsn ready for review when you hve time 🙏

sentry-samples/kmp-app-spm/androidApp/build.gradle.kts
```diff
@@ -7,15 +7,21 @@ plugins {
 }
 
 java {
```
romtsn, line 9: same with the toolchain

romtsn (approved):

buenaflor:
Just noticed that this messes up publication as K2 produces the output of .klib files a bit differently. will adjust the PR

PR #414, opened 2025-07-04
ci(publishing): Use shim ossrh-staging-api server to publish to the new central repo 

romtsn:
_#skip-changelog_

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -6,8 +6,8 @@ targets:
     gradleCliPath: ./gradlew
     mavenCliPath: scripts/mvnw
     mavenSettingsPath: scripts/settings.xml
-    mavenRepoId: ossrh
-    mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
+    mavenRepoId: ossrh-staging-api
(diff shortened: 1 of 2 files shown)


.craft.yml
```diff
@@ -21,13 +21,13 @@ targets:
...
+    mavenRepoId: ossrh-staging-api
+    mavenRepoUrl: https://ossrh-staging-api.central.sonatype.com/service/local/staging/deploy/maven2/
     android: false
     includeNames: /sentry-kotlin-multiplatform-gradle-plugin.*$/
   - name: github
   - name: registry
     sdks:
       maven:io.sentry:sentry-kotlin-multiplatform:
-#      maven:io.sentry:sentry-kotlin-multiplatform-gradle-plugin:
+      maven:io.sentry:sentry-kotlin-multiplatform-gradle-plugin:
```
romtsn, line 32: we forgot to uncomment it after the first release right? 🙈

buenaflor: oopsie, I dont think it mattered much though since its synced with the sdk versions

but yeah we should uncomment it

PR #425, opened 2025-07-17
Update `distZip` task to use the locally published artifacts via `publishToMavenLocal`

buenaflor:
Note: this branch merges into the Kotlin `2.1.21` bump branch since the kotlin bump doesn't work with the current/old  publication anyway

Instead of processing artifacts ourselves let's just use the locally published artifacts directly which makes it more robust for the future

CI test

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/.github/workflows/upload-artifacts.yml
+++ b/.github/workflows/upload-artifacts.yml
@@ -1,6 +1,8 @@
 name: "Generate and upload distributions"
 on:
-  pull_request:
+  push:
+    branches:
+      - release/**
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/upload-artifacts.yml
+++ b/.github/workflows/upload-artifacts.yml
@@ -1,8 +1,6 @@
 name: "Generate and upload distributions"
 on:
-  push:
-    branches:
-      - release/**
+  pull_request:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -35,15 +35,15 @@ subprojects {
         val buildPublishDir = "${project.layout.buildDirectory.get().asFile.path}${sep}sentry-local-publish$sep"
 
         configure<DistributionContainer> {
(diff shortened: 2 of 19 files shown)


.github/workflows/upload-artifacts.yml
```diff
@@ -1,8 +1,6 @@
 name: "Generate and upload distributions"
 on:
-  push:
-    branches:
-      - release/**
```
buenaflor, line 5: enabled temporarily for testing

buenaflor: Works well, the distribution validation task also passed which means it most likely looks good

https://github.com/getsentry/sentry-kotlin-multiplatform/actions/runs/16355042742

build.gradle.kts
```diff
@@ -31,16 +31,19 @@ subprojects {
...
 
         configure<DistributionContainer> {
-            this.configureForMultiplatform(this@subprojects)
+            configureForMultiplatform(this@subprojects, buildPublishDir)
         }
 
         tasks.named("distZip").configure {
-            this.dependsOn("publishToMavenLocal")
-            this.doLast {
+            System.setProperty("maven.repo.local", buildPublishDir)
```
romtsn, line 42: so this will tell the task where to publish?

buenaflor: yep, it then publishes into the specified dir

buildSrc/src/main/java/Publication.kt
```diff
@@ -1,228 +1,47 @@
...
-private fun CopySpec.renameModule(projectName: String, renameTo: String = "", version: String) {
-    var target = ""
-    if (renameTo.isNotEmpty()) {
-        target = "-$renameTo"
-    }
-    rename {
-        it.replace("module.json", "$projectName$target-$version.module")
-    }
-}
+            // Rename the POM since craft looks for pom-default.xml
```
romtsn, line 37: ugh, I guess ideal-world we wouldn't need this, but would have to change craft again 🙈 maybe makes sense to align this between repos and change craft at some point

Issue #77, opened 2023-04-19
Stacktrace caught on Apple targets include unnecessary frames


buenaflor:
### Steps to Reproduce

Run the iOS sentry sample app (with dsyms uploading configured) and click on the Capture Exception button

### Expected Result

Stacktrace would cut off at frame `kfun:sample.kmp.app.LoginImpl#login{}`

### Actual Result

<img width="840" alt="Screenshot 2023-04-19 at 14 45 13" src="https://user-images.githubusercontent.com/23364143/233078713-a8305ab3-42d6-4337-a8f5-f26bcbee62ce.png">

buenaflor:
After investigation, we need to either remove these frames or set them as non-in-app in the backend.

buenaflor:
Need to re-evaluate if this is also happening in symbolicated stacktraces, the example shows a debug build not a release build

PR #493, opened 2025-12-03
Improve stacktrace of manually captured exceptions on iOS

### LINKED ISSUE #341: captureException captures the wrong stack trace
### Platform

Apple

### Installed

CocoaPods

### Version

0.9.0

### Steps to Reproduce

1. create a `Throwable` at site `A`
2. call `captureException()` from site `B`

### Expected Result

Exception on the dashboard should include the stack trace at the time the exception was instantiated (i.e. site `A`)

### Actual Result

Exception on the dashboard includes the stack trace at the time `captureException()` is called (i.e. site `B`)

Image

Notice how the top of the stack includes Sentry. Breadcrumbs seem accurate though. 

Works as expected on Android, but not on iOS.

buenaflor:
**Note: this will mess up issue grouping because it changes the stacktrace**

---

## :scroll: Description

Improve stacktrace of manually captured exceptions on iOS - unhandled exceptions already work well

On iOS, when calling `captureException()` with a Kotlin `Throwable`, the stack trace shows where `captureException()` was **called**, not where the exception was **instantiated**.

### Example
```kotlin
try {
    UserApi.fetchUser()  // Exception thrown HERE
} catch (e: Exception) {
    ErrorHandler.report(e)  // captureException called HERE
}
```

**Expected stack trace:** `UserApi.fetchUser()` → ...  
**Actual stack trace:** `ErrorHandler.report()` → ...

This works correctly on Android, but not on iOS.

### Cause
`SentrySDK.captureException()` in Sentry Cocoa ignores the `NSException`'s `callStackReturnAddresses()` and captures a fresh stack trace at call time.

### Fix
Instead of calling `SentrySDK.captureException()`, we now:
1. Create a `SentryEvent` directly
2. Extract the correct stack trace from the `NSException` using `NSExceptionKt_SentryCrashStackCursorFromNSException()`
3. Send via `SentrySDK.captureEvent()`

This is the same approach already used for unhandled exceptions in `SentryUnhandledExceptions.kt` (the old way at least).

Fixes #341 and https://github.com/getsentry/sentry-kotlin-multiplatform/issues/77

Currently only manually. I tried running integration and e2e tests but the issue is that the frames are obfuscated so I need to find a way to do an e2e test where dsyms are also uploaded - probably run through one of the examples

**Before:**
Screenshot 2025-12-03 at 21 41 22

---

**After:**
Screenshot 2025-12-03 at 21 49 57

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,17 @@
 # Changelog
 
-## 0.22.0
+## Unreleased
+
+⚠️ This release will affect issue grouping for iOS events as Sentry now captures correct stacktraces for manually captured and crashed iOS events.
 
 ### Features
 
 - Improve iOS crash reports by adding scope data ([#491](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/491))
- - ⚠️ This change will most likely affect issue grouping as Sentry now properly symbolicates Kotlin iOS crashes
+- Improve stacktrace of manually captured exceptions on iOS ([#493](https://github.com/getsentry/sentry-kotlin-multiplatform/pull/493))
+
+## 0.22.0
+
+### Features
 
 ### Dependencies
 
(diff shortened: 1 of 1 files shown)


romtsn (approved):
Nice work!

PR #556, opened 2026-06-09
[1/4] chore(build): Bump Kotlin to 2.2.21 and Gradle to 8.13 to unblock spm4Kmp

buenaflor:
Upgrade to Kotlin 2.2.21, Gradle 8.13, and AGP 8.9.1 for spm4Kmp and Xcode 26.

- Set `compileSdk` and sample `targetSdk` to 36, matching Sentry Android 8.41.0; retain `minSdk` 21.
- Adapt build/publishing configuration and preserve the Android `BuildConfig` API.
- Exclude unsupported stub tests and update detekt configuration. Formatting migration is in #561.

Validation: formatting, detekt, Android/JVM API checks, JVM/Android/iOS/macOS tests, and both Android release samples with R8 passed. The local full CocoaPods build is blocked by a missing tvOS destination; Apple CI passed.


kotlin-js-store/yarn.lock
```diff
@@ -0,0 +1,2168 @@
...
+jest-worker@^27.4.5:
+  version "27.5.1"
+  resolved "https://registry.yarnpkg.com/jest-worker/-/jest-worker-27.5.1.tgz#8d146f0900e8973b106b6f73cc1e9a8cb86f8db0"
+  integrity sha512-7vuh85V5cdDofPyxn58nrPjBktZo0u9x1g8WtjQol+jZDaE+fhN+cIvTj11GndBnMnyfrUOG1sZQxCdjKh+DKg==
+  dependencies:
+    "@types/node" "*"
+    merge-stream "^2.0.0"
+    supports-color "^8.0.0"
+
+js-yaml@^4.1.0:
```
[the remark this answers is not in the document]
buenaflor: Resolved in 019eb07 by removing the generated Yarn lockfile. `kotlinUpgradeYarnLock` reports that this project has no NPM dependencies and deletes the empty lockfile; `jsBrowserTest` succeeds without it. The fresh Semgrep scan passes.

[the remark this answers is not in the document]
buenaflor: Resolved in 019eb07 by removing the generated Yarn lockfile. `kotlinUpgradeYarnLock` reports that this project has no NPM dependencies and deletes the empty lockfile; `jsBrowserTest` succeeds without it. The fresh Semgrep scan passes.

sentry-kotlin-multiplatform/build.gradle.kts
```diff
@@ -18,15 +19,14 @@ plugins {
...
-    defaults {
-        // adds the contents of the reports of `release` Android build variant to default reports
-        mergeWith("release")
-    }
-}
-
 android {
+    namespace = "io.sentry.kotlin.multiplatform"
     compileSdk = Config.Android.compileSdkVersion
+    // AGP 8 disables BuildConfig generation by default; keep it on to preserve
```
markushi, line 25: 💯

buildSrc/src/main/java/Config.kt
```diff
@@ -1,8 +1,8 @@
 object Config {
-    val agpVersion = "7.4.2"
-    val kotlinVersion = "2.1.21"
-    val composePluginVersion = "1.8.0"
-    val gradleMavenPublishPluginVersion = "0.18.0"
+    val agpVersion = "8.9.1"
+    val kotlinVersion = "2.2.21"
```
markushi, line 3: We're implicitly bumping the Kotlin language version here too, which could cause issues with (E.g. app uses Kotlin 2.1.21, but SDK was compiled against 2.2.21). Maybe worth double checking if we should configure a `languageVersion`, see https://blog.gradle.org/three-kotlin-versions-in-a-gradle-project#2-the-kotlin-language-version-for-your-code

markushi (approved):
Looking good!

buildSrc/src/main/java/Config.kt
buenaflor: unfortunately setting both to 2.1 does not make a Native library built with compiler 2.2.21 consumable by compiler 2.1.21

it would benefit in JVM/Android only cases, but KMP use cases are rarely only those. it's almost always in combination with a native target like iOS

PR #567, opened 2026-09-10
feat: Update Sentry Cocoa from `8.58.2` to `9.28.0` via spm4kmp

buenaflor:
Upgrade Sentry Cocoa from 8.58.2 to 9.28.0 and adapt options, logs, feedback, and crash handling to its APIs.

- Use `SentrySDK.internal` through a Swift adapter for SDK metadata, envelope storage, and debug images. Remove obsolete private declarations; the remaining declarations match Cocoa 9.28.0.
- Update framework selection and the watchOS simulator workaround.
- Preserve fatal-event sampling/filtering, native `beforeSend` replacements, and common user-feedback behavior.

Compatibility: requires iOS/tvOS 15, macOS 12, and watchOS 9. `watchosArm32` becomes a no-op SDK. CocoaPods is unsupported; use spm4Kmp. The `cocoapods.Sentry` import namespace is preserved.

Validation: formatting, SDK/plugin detekt, full build, API checks, and all 144 Gradle plugin tests passed. Apple integration tests cover exception frames/debug images, sampling, scope enrichment, filtering, and C++ monitor configuration across SDK restarts. Local validation excludes service E2E tests and physical-device/x64 execution.


sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/CocoaFrameworkLinker.kt
```diff
@@ -17,6 +18,10 @@ class CocoaFrameworkLinker(
 ) {
     fun configure(appleTargets: List<KotlinNativeTarget>) {
         appleTargets.forEach { target ->
+            if (target.konanTarget == KonanTarget.WATCHOS_ARM32) {
+                logger.info("Skipping Sentry Cocoa linking for no-op target ${target.name}.")
+                return@forEach
+            }
```
markushi: `nit`: Maybe I'm still doing too much Java, but this feels easier to read:
```suggestion
    fun configure(appleTargets: List<KotlinNativeTarget>) {
        for (target in appleTargets) {
            if (target.konanTarget == KonanTarget.WATCHOS_ARM32) {
                logger.info("Skipping Sentry Cocoa linking for no-op target ${target.name}.")
                continue
            }
```

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/CocoaFrameworkLinker.kt
```diff
@@ -17,6 +18,10 @@ class CocoaFrameworkLinker(
 ) {
     fun configure(appleTargets: List<KotlinNativeTarget>) {
         appleTargets.forEach { target ->
+            if (target.konanTarget == KonanTarget.WATCHOS_ARM32) {
+                logger.info("Skipping Sentry Cocoa linking for no-op target ${target.name}.")
```
markushi: what's the rationale here? Is WATCHOS_ARM32 simply unsupported by kmp, our sdk or the cocoa SDK?

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/Spm4KmpWatchosSimulator.kt
```diff
@@ -0,0 +1,56 @@
...
+
+private const val COPY_TASK_NAME = "copySentrySpmWatchosSimulatorFramework"
+private const val SPM_TASK_PREFIX = "SwiftPackageConfigAppleSentryCocoa"
+
+/**
+ * spm4Kmp 1.9.5 uses aarch64 for the watchOS simulator triple, which SwiftPM does not match to
+ * the arm64 binary slice. Link-only consumers still need that framework beside the bridge archive.
+ * Called only for packages owned by Sentry auto-installation.
+ */
+internal fun Project.registerWatchosSimulatorFrameworkCopy() {
```
markushi: these tasks are all flavor-less right? Like there's no debug/release variant or any custom build flavor we'd need to watch out for?

sentry-kotlin-multiplatform-gradle-plugin/src/test/java/io/sentry/kotlin/multiplatform/gradle/SentryFrameworkArchitectureTest.kt
```diff
@@ -19,18 +21,17 @@ class SentryFrameworkArchitectureTest {
...
         fun cocoaVersions(): List<Arguments> =
-            listOf(
-                Arguments.of("8.37.0"),
-                Arguments.of("8.38.0"),
-                Arguments.of("8.58.2"),
-//            Arguments.of("latest"),
-                // TODO: Latest is already v9 which is currently failing - let's fix this when we bump to v9
-            )
+            listOf("8.37.0", "8.38.0", "8.58.2", "9.28.0").flatMap { version ->
+                listOf(true, false).map { isStatic -> Arguments.of(version, isStatic) }
```
markushi: `nit`, at least that would be easier for my brain to understand 😅  Feel free to keep it as is.
```suggestion
                listOf(Arguments.of(version, true), Arguments.of(version, false))
```

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/nsexception/UnhandledExceptionHook.kt
```diff
@@ -16,6 +16,8 @@ package io.sentry.kotlin.multiplatform.nsexception
 
 import kotlin.concurrent.AtomicReference
 
+private val installedSentryHook = AtomicReference<ReportUnhandledExceptionHook?>(null)
```
markushi: I'd add a lock here to avoid any threading issues
```suggestion
private val installedSentryHook = AtomicReference<ReportUnhandledExceptionHook?>(null)
private val installedSentryHookLock = Lock()
```

then within wrapUnhandledExceptionHook, you could wrap the code block with ` installedSentryHookLock.withLock { ... }`

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryBridge.apple.kt
```diff
@@ -23,34 +36,44 @@ public actual abstract class Context
...
-        // Return early if the user's beforeSend returns null
-        if (userDefinedBeforeSend != null && userDefinedBeforeSend.invoke(event) == null) {
-            return@beforeSend null
-        }
+        // A replacement event could erase the marker on a duplicate Kotlin termination report.
+        // Suppress that report before invoking the user's callback.
+        if (dropKotlinCrashEvent(event) == null) return@beforeSend null
+        val processedEvent =
+            if (userDefinedBeforeSend != null) {
+                userDefinedBeforeSend.invoke(event) ?: return@beforeSend null
```
markushi: Not sure if it's relevant for this PR, but user defined methods should by try/catched.

sentry-kotlin-multiplatform/src/appleTest/kotlin/io/sentry/kotlin/multiplatform/SentryEventConfigurator.kt
```diff
@@ -15,7 +15,7 @@ actual class SentryEventConfigurator {
     actual fun applyOptions(options: SentryOptions): SentryEvent? {
         val cocoaOptions = CocoaSentryOptions()
         cocoaOptions.applyCocoaBaseOptions(options)
-        val cocoaModifiedSentryEvent = cocoaOptions.beforeSend?.invoke(cocoaSentryEvent)
+        val cocoaModifiedSentryEvent = cocoaOptions.beforeSend()?.invoke(cocoaSentryEvent)
```
markushi: again, (and probably out of scope), I guess those callbacks should be guarded by try catch

sentry-kotlin-multiplatform/src/nativeInterop/cinterop/SentryInternal/SentryCrashMonitor_NSException+NSExceptionKt.h
```diff
@@ -15,18 +15,23 @@
...
+#import <SentryStacktraceBuilder.h>
+#include <stdlib.h>
+#include <assert.h>
 
 // Similar to how Sentry converts stacktraces from NSExceptions
 // https://github.com/getsentry/sentry-cocoa/blob/167de8bea5a0effef3aaa5c99c540088de30b361/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_NSException.m#L60
-SentryCrashStackCursor NSExceptionKt_SentryCrashStackCursorFromNSException(NSException *exception) {
+SentryStacktrace *NSExceptionKt_SentryStacktraceFromNSException(SentryStacktraceBuilder *builder, NSException *exception) {
     NSArray *addresses = [exception callStackReturnAddresses];
     NSUInteger numFrames = addresses.count;
```
markushi: could we just return null here? this would avoid the `MAX(numFrames, 1)` workaround.
```suggestion
    NSUInteger numFrames = addresses.count;
    if (numFrames == 0) {
        return NULL;
    }
```

sentry-kotlin-multiplatform/src/nativeInterop/cinterop/SentryInternal/SentryCrashMonitor_NSException+NSExceptionKt.h
```diff
@@ -15,18 +15,23 @@
...
-    uintptr_t *callstack = malloc(numFrames * sizeof(*callstack));
+    uintptr_t *callstack = malloc(MAX(numFrames, 1) * sizeof(*callstack));
     assert(callstack != NULL);
```
markushi: If malloc fails, I'd also just return null instead
```suggestion
    if (callstack == NULL) {
        return null;
    }
```

markushi (commented):
Nice job, looking pretty good already - I left a few comments! On top of that, here's some more unfiltered feedback from my agent run below, maybe the first two are worth investigating in more detail.

- User feedback has a new shape. It is not only a rename: feedback is now a separate item with its own event id, linked with associatedEventId. Ingestion and alerting change. 
The CHANGELOG puts this under "Fixes" and does not say this.

- Fatal events now get the full scope path. prepareEvent(..., isFatalEvent = false) adds scope plus device/culture/view contexts, 
but it skips the onCrashedLastRun block. So onCrashedLastRun can stop to operate for Kotlin fatals. 
This is the most important runtime change, and it is not in the CHANGELOG.

- The klib is pinned to Cocoa 9.28.0 through Swift @_spi(Private) symbols. If a user sets autoInstall.spm.sentryCocoaVersion to a different 9.x, link errors are possible. There is no version check, and no warning in the README.

- Two private declarations carry load: retrieveStacktraceFromCursor: (no header, SDK_V10 churn around it) and SentrySDK.startOption (loses @objc under SDK_V10). Both will break on Cocoa 10 with no compile-time signal.

- CocoaPods leftovers stay on disk: sentry-samples/kmp-app-cocoapods/**, .run/*-cocoapods.run.xml, and a shared.podspec still at 8.58.2.

- Spm4KmpIntegration.kt — add a check of the configured sentryCocoaVersion against BuildKonfig.SENTRY_COCOA_VERSION; fail or warn on a difference.

- Spm4KmpIntegration.kt — minimumDeploymentVersion uses toInt(). A target like "15.0.1" throws NumberFormatException from our plugin. Use toIntOrNull().

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/CocoaFrameworkLinker.kt
buenaflor: it's unsupported by the Cocoa SDK since `9.26.0` (or `9.28.0`)

sentry-kotlin-multiplatform/src/appleMain/kotlin/io/sentry/kotlin/multiplatform/SentryBridge.apple.kt
buenaflor: I'll address this in another PR

sentry-kotlin-multiplatform/src/appleTest/kotlin/io/sentry/kotlin/multiplatform/SentryEventConfigurator.kt
buenaflor: I'll address this in another PR

sentry-kotlin-multiplatform-gradle-plugin/src/main/java/io/sentry/kotlin/multiplatform/gradle/Spm4KmpWatchosSimulator.kt
buenaflor: yup, no need to watch out for flavours

buenaflor:
@markushi thx! addressed the inline suggestions and removed the retired CocoaPods sample. The changelog now clarifies the feedback event’s separate ID.

The fatal-event change only affects the fallback when Cocoa’s native handler is unavailable (which it shouldn't in most cases) - I added a comment explaining the callback limitation.

I also added a Cocoa version-mismatch warning locally. The 15.0.1 parsing concern also is a false positive

I think it should be good to go now

